PR #122, opened 2023-04-27
Konvent

juro-ub:
Includes:
- SQL for Konvent
- Datamodel
- Logic
- View


updates/12.sql
```diff
@@ -0,0 +1,17 @@
...
+  `Bezeichnung` VARCHAR(50) CHARACTER SET 'utf8mb4' COLLATE 'utf8mb4_unicode_ci' NOT NULL,
+  PRIMARY KEY (`ID`),
+  UNIQUE (`Bezeichnung`)
+  )
+ENGINE = InnoDB
+DEFAULT CHARACTER SET = utf8mb4
+COLLATE = utf8mb4_unicode_ci;
+
+/*Insert a default value*/
+INSERT INTO `neg`.`selektion_konvent` (Bezeichnung) VALUES ('-');
```
mtrojan-ub, line 12: I see two options here:
- Either we avoid using an explicit default value and set the references from other tables to NULL in that case (would be the best DB design). However this might cause problems with certain save forms etc.
- Or, instead of INT UNSIGNED we just use INT and explicitly set the ID of the default value to -1, which is at least similar to most other selektion_- tables.

juro-ub: I would try to prevent using negative values in the primary key columns.
I have adjusted the dosave.jsp file to handle NULL values from the formular,
so we have a clean db design for this

juro-ub: As we discussed i will change the column definition from INT UNSIGNED to INT and will insert a default value with the id -1.

PR #135, opened 2023-06-22
Set Auto Increment

juro-ub:
For the Alignment Tool it is necessary that all ID columns are set to auto_increment


updates/13.sql
```diff
@@ -0,0 +1,7 @@
+--avoid foreign key check for this session, otherwise the alter table statement will not work
+SET FOREIGN_KEY_CHECKS = 0;
```
mtrojan-ub, line 2: Disabling foreign key checks is a bad thing. The foreign keys are guaranteeing the consistency of the database.

If you add auto_increment on an existing column, the IDs will be reindexed. So if there are deleted values, the ids will no longer match the foreign keys.

For example, we have 11 Dichtung and 13 Testament, so 12 is a gap which has been deleted. If you add auto_increment now, 13 will be reindexed as 12 which means all foreign references will get broken.

So there must be a better way to achieve this. I found a stackoverflow article on this:
https://stackoverflow.com/questions/740358/reorder-reset-auto-increment-primary-key

Maybe the AUTO_INCREMENT value needs to be set explicitly to the max current value of the database?

mtrojan-ub (changes requested):

updates/13.sql
juro-ub: I have changed it as suggested in the post of stackoverflow.
But there is still an error message:
`Error Code: 1451. Cannot delete or update a parent row: a foreign key constraint fails (`neg`.`einzelbeleg`, CONSTRAINT `einzelbeleg_QuelleGattungID` FOREIGN KEY (`QuelleGattungID`) REFERENCES `selektion_quellengattung` (`ID`))
`
The following statement return 0 rows. So the column QuelleGattungID of einzelbelg seems to be fine:
`SELECT ID FROM einzelbeleg WHERE QuelleGattungID NOT IN (SELECT `ID` FROM `selektion_quellengattung`);`

Not sure how to solve it

mtrojan-ub: This SELECT will only prove that all IDs still exist in the other table, but not that the actual values are the same as before the operation. But you're right, this is indeed a hard nut to crack. I will have another look, but it might take some time.

mtrojan-ub: After some more research i found this article:
https://stackoverflow.com/questions/50117614/auto-increment-primary-key-and-update-its-references-in-other-tables-at-the-same

So maybe we need to delete the existing foreign key & recreate it using ON UPDATE CASCADE before we add the AUTO INCREMENT to the corresponding column?

juro-ub: Yes. It works now

updates/13.sql
```diff
@@ -0,0 +1,4 @@
+SET SESSION sql_mode='NO_AUTO_VALUE_ON_ZERO';
+ALTER TABLE einzelbeleg DROP FOREIGN KEY einzelbeleg_QuelleGattungID;
+ALTER TABLE selektion_quellengattung MODIFY COLUMN ID INT NOT NULL auto_increment;
+ALTER TABLE einzelbeleg ADD CONSTRAINT einzelbeleg_QuelleGattungID FOREIGN KEY (QuelleGattungID) REFERENCES selektion_quellengattung(ID) ON UPDATE CASCADE;
```
mtrojan-ub, line 4: I think the foreign key must be restored before the auto increment is set else the references will break during migration

updates/13.sql
```diff
@@ -0,0 +1,4 @@
+SET SESSION sql_mode='NO_AUTO_VALUE_ON_ZERO';
```
mtrojan-ub, line 1: can we remove the NO_AUTO_VALUE_ON_ZERO change?

updates/13.sql
juro-ub: I have changed the order

updates/13.sql
juro-ub: I think we need it in this case, because the id column contains '0' values. 
Otherwise an error is thrown.

https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html#sqlmode_no_auto_value_on_zero

mtrojan-ub: As far as i understand the SQL mode, it can consist of several settings at the same time. The default depends on the server settings.

E.g. on my machine this is a special value which is documented in the project's `README.md`: `STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION`

If you use the `SET SESSION` command, you will set the value to  `NO_AUTO_VALUE_ON_ZERO`, but therefore lose all other existing changes like e.g. `STRICT_TRANS_TABLES` which might as well be necessary for the migration.
The other aspect is that we might execute multiple updates during the same session. So you should do the following

1) Find a way to add this setting to the existing ones, instead of entirely overriding it
2) Save the previous value in a variable and restore it after the migration

juro-ub: I have found a way to restore the sql_mode

```
-- restore sql mode
SET SESSION sql_mode=default;
```

mtrojan-ub: some more changes please:
- Changing the FOREIGN KEY doesnt require the sql_mode to be changed, so please move the first SET SESSION sql_mode statement directly above the introduction of AUTO_INCREMENT
- AUTO_INCREMENT is a sql keyword and should be written in uppercase
- 13.sql already existed and contained a change which had nothing to do with the AUTO_INCREMENT change. Moreover there have been additional .sql files created in the meantime and already been applied on certain machines. Please move the changes related to selektion_quellengattung and the FOREIGN KEY to a new separate file (i think 17.sql should the next non-existing number in the sequence.)

juro-ub: 1) moved to set auto_increment
2)fixed
3)it is now in 17.sql

mtrojan-ub: This aspect is still not solved:

> Find a way to add this setting (NO_AUTO_VALUE_ON_ZERO) to the existing ones, instead of entirely overriding it

juro-ub: Now the CONCAT Sql function will be used to append the sql_mode to the existing ones

updates/17.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE einzelbeleg DROP FOREIGN KEY einzelbeleg_QuelleGattungID;
+-- this is necessary because the id column contains 0 values. Otherwise an error appears
+ALTER TABLE einzelbeleg ADD CONSTRAINT einzelbeleg_QuelleGattungID FOREIGN KEY (QuelleGattungID) REFERENCES selektion_quellengattung(ID) ON UPDATE CASCADE;
+SET SESSION sql_mode = CONCAT(@@SQL_MODE, ',NO_AUTO_VALUE_ON_ZERO');
+ALTER TABLE selektion_quellengattung MODIFY COLUMN ID INT NOT NULL AUTO_INCREMENT;
```
mtrojan-ub: The sql mode seems to be OK now, but if i execute all statements in the given order on my machine it fails:
`Cannot change column 'ID': used in a foreign key constraint 'einzelbeleg_QuelleGattungID' of table 'neg.einzelbeleg'`

updates/22.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE einzelbeleg DROP FOREIGN KEY einzelbeleg_QuelleGattungID;
+-- this is necessary because the id column contains 0 values. Otherwise an error appears
+ALTER TABLE einzelbeleg ADD CONSTRAINT einzelbeleg_QuelleGattungID FOREIGN KEY (QuelleGattungID) REFERENCES selektion_quellengattung(ID) ON UPDATE CASCADE;
```
mtrojan-ub: In the meantime, another Foreign key has been added for hierarchies which needs to be considered in this update as well => selektion_quellengattung.parentId is referencing selektion_quellengattung.ID

juro-ub: I have added the foreign key to the script

updates/17.sql
juro-ub: I had problems to reproduce the issue on my machine.
But i have changed the order of the statements.
On my machine it is working.
Can you please try it again

juro-ub:
I have tested the script with the actual dump from the testserver.
On my Machine it is working now.
Can you please test and review again @mtrojan-ub

mtrojan-ub:
Looks good now, thanks!

PR #141, opened 2023-07-11
Traceability of Migrated Data

juro-ub:
Add cols to neg database tables.
(Not all neg tables should have provenance information. Will be discussed in next workshop)
- The column provenance will be used to associate migrated data with a source system (batch import or dmp)
- The column external_dmp_id will be used to store the dmp primary key of a migrated row


updates/14.sql
```diff
@@ -0,0 +1,159 @@
+#provenance
+#
+#add a enum to associated migrated data with a source system (a source system can be for example dmp or batch import)
+#
+ALTER TABLE bemerkung ADD provenance VARCHAR (255);
```
mtrojan-ub: If we add multiple provenance-based fields, we should rename them:
- provenance => provenance_source, Default should be something like 'NeG' or 'neg'
- external_dmp_id => provenance_id (in the long term this could be something else than dmp, so we should not include the database name in the field), DEFAULT NULL (in case it is NeG we dont store any external ids here)

default values are important, else we cannot add new values anymore without changing lots of code

juro-ub: I have implemented the changes

juro-ub:
@mtrojan-ub 
To add the provenience information in the Import program this PR needs to be merged, to add the missing columns.
Could you please review it

updates/julian_1.sql
```diff
@@ -0,0 +1,159 @@
+#provenance_source
+#
+#add a enum to associated migrated data with a source system (a source system can be for example dmp or batch import)
+#
+ALTER TABLE bemerkung ADD provenance_source VARCHAR (255) NOT NULL DEFAULT('NeG');
+ALTER TABLE benutzer ADD provenance_source VARCHAR (255) NOT NULL DEFAULT('NeG');
+ALTER TABLE benutzer_gruppe ADD provenance_source VARCHAR (255) NOT NULL DEFAULT('NeG');
+ALTER TABLE datenbank_filter ADD provenance_source VARCHAR (255) NOT NULL DEFAULT('NeG');
```
mtrojan-ub: There are certain tables that are related to the internal system and will never be synchronized from other systems, so we should not add provenance information. For example:
- benutzer
- benutzer_gruppe
- datenbank_filter
- datenbank_mapping
- datenbank_selektion
- datenbank_sprachen
- datenbank_texte
- suche_favoriten
- content
- (maybe more? please check again.)

updates/julian_1.sql
```diff
@@ -0,0 +1,159 @@
...
+ALTER TABLE urkunde_betreff ADD provenance_source VARCHAR (255) NOT NULL DEFAULT('NeG');
+ALTER TABLE urkunde_dorsalnotiz ADD provenance_source VARCHAR (255) NOT NULL DEFAULT('NeG');
+ALTER TABLE urkunde_hataussteller ADD provenance_source VARCHAR (255) NOT NULL DEFAULT('NeG');
+ALTER TABLE urkunde_hatempfaenger ADD provenance_source VARCHAR (255) NOT NULL DEFAULT('NeG');
+#
+#
+#add a field to neg tables to associated migrated rows with dmp rows
+#
+#
+ALTER TABLE bemerkung ADD provenance_id VARCHAR (255) DEFAULT NULL;
```
mtrojan-ub: I think we should also add a UNIQUE constraint to every table here, for several reasons:
- consistency during execution of our import program (each ID should only exist once)
- Performance improvement when doing lookups

updates/julian_1.sql
mtrojan-ub: Note: There might be cases where we do not have a separate ID. E.g. in DMP there is a field namen.funktion which belongs to the namen table but will be mapped to the separate NeG.einzelbeleg_hatfunktion table, so you technically do not have an ID. In this case we should set the provenance_source to DMP but we cannot set a specific provenance_id. However i won't mind adding the provenance_id field to the schema in this case to keep the design more straightforward.

updates/julian_1.sql
juro-ub: fixed!

updates/julian_1.sql
juro-ub: System tables are removed now

PR #145, opened 2023-07-14
Fix HTML entity escaping in noarray.select.jsp

mtrojan-ub:
.. problem is that we have to call DBtoHTML **BEFORE** calling format() else it will destroy the inserted entities.


juro-ub (approved):
Works fine

rnaiser (approved):
good !!

PR #148, opened 2023-07-17
Multiple occasions: Use hibernate instead of deprecated direct Connections

mtrojan-ub:


src/main/webapp/forms/counter.jsp
```diff
@@ -1,63 +1,39 @@
...
     }
 
+    String sql = DatenbankDB.getFilterSql(guest + title, filter);
     sql = sql.replace("*", "count(*) c");
     sql = sql.replace("###", filterParameter);
 
-    rs = st.executeQuery(sql+(sql.contains("WHERE")?" AND ":" WHERE ")+title+".ID < "+id);
-    if (rs.next())
-      akt = rs.getInt("c") + 1;
+    Object[] columns = AbstractBase.getRowNative(sql + (sql.contains("WHERE") ? " AND " : " WHERE ") + title + ".ID < " + id);
```
juro-ub: In my built of this branch i got an execption on line 28.

`Caused by: java.lang.ClassCastException: class java.math.BigInteger cannot be cast to class [Ljava.lang.Object; (java.math.BigInteger and [Ljava.lang.Object; are in module java.base of loader 'bootstrap')`

pom.xml
```diff
@@ -167,6 +167,24 @@
                 <artifactId>maven-failsafe-plugin</artifactId>
                 <version>2.22.2</version>
             </plugin>
+
+            <!-- Download javadoc for all dependencies -->
```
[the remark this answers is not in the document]
mtrojan-ub: That's true, but this is only a 1-time-action that will not be executed every build (only if a version is changed or new dependency is added). I found it quite useful to access the direct documentation e.g. of hibernate classes.

src/main/webapp/forms/counter.jsp
mtrojan-ub: solved

rnaiser (approved):

PR #149, opened 2023-07-25
Textkritische Anmerkung

juro-ub:
New Field "Textkritische Anmerkung"
image


mtrojan-ub:
As far as i remember, the main task was to add the new field, which means that the existing field + also its name should stay the same. The new field should be named "Textkritische Anmerkung". 

Moreover, the new field is related to "kritik" in DMP. For consistency reasons it might be better to change the name of the new table to "selektion_kritik". This might also be useful since it is linked to multiple entities in DMP (PN, Titel, Datum) and we can later reference it properly from multiple tables in NeG as well if necessary. "Textkritische Anmerkung" will only be the reference from "Einzelbeleg".

juro-ub:
> As far as i remember, the main task was to add the new field, which means that the existing field + also its name should stay the same. The new field should be named "Textkritische Anmerkung".
> 
> Moreover, the new field is related to "kritik" in DMP. For consistency reasons it might be better to change the name of the new table to "selektion_kritik". This might also be useful since it is linked to multiple entities in DMP (PN, Titel, Datum) and we can later reference it properly from multiple tables in NeG as well if necessary. "Textkritische Anmerkung" will only be the reference from "Einzelbeleg".

The purpose of this PR is now to add only the new field "Textkritische Anmerkung".
I have reverted the changes to the existing field Kontext (s. Screenshot).
I have renamed the table selektion_kontext to selektion_kritik and also the variables

image

mtrojan-ub:
This looks good and i'd really like to merge it today! Can you please solve the conflict by renaming 22.sql to 23.sql?

juro-ub:
> This looks good and i'd really like to merge it today! Can you please solve the conflict by renaming 22.sql to 23.sql?

I have renamed the file

PR #154, opened 2023-07-26
Search-related changes

rnaiser:


src/main/java/de/uni_tuebingen/ub/nppm/servlet/backend/ContentServlet.java
```diff
@@ -20,35 +19,51 @@
...
-        // TinyMCE anzeigen
+        // show TinyMCE
         if (request.getParameter("loadFile") != null) {
             RequestDispatcher rd = request.getRequestDispatcher("tinyMce.jsp");
             rd.include(request, response);
         } else {
-            // Aktionen:
+        //show fileManagement
+
+            out.println("<h2>Html Dateien & Bilder verwalten</h2>");
```
mtrojan-ub: This HTML output should typically be moved to the corresponding JSP template

rnaiser:
bei erweiterte suche , Ausgabe --> wenn man Ämter und Weihegrade der Person ohne inklusive Zeitraum suchen lies gab es keine ausgabe, es ging nur mit inklusive Zeitraum, das ist nun verbessert worden und funktioniert

mtrojan-ub:
As discussed: Since this PR is already open for a while, i will manually merge the content / fileManagement-related changes in a different PR to develop, but we're gonna leave the search related changes open.

src/main/java/de/uni_tuebingen/ub/nppm/db/FrontendExtendedSearch.java
```diff
@@ -11,29 +11,50 @@
...
                     }
-                    if (row[i] != null) {
-                        fieldVal.put(fields[i].trim(), row[i].toString());
+                    ret.add(fieldVal);
+                }//end for (Object[] row : rows)
+
+                return ret;
+            }//end fields.length
+            else {
+                NativeQuery sqlQuery = session.createNativeQuery(sql);
```
mtrojan-ub, line 40: This line as well as the initialization of the ret variable are the same in both blocks, so you might actually move them before the if...else

mtrojan-ub:
There seem to be merge conflicts with freie_suche.jsp

PR #171, opened 2023-08-30
inc.modul.jsp refactoring (use DB class instead of direct DB connection)

mtrojan-ub:


mtrojan-ub:
This PR will be closed since the work was pretty outdated + has already been continued in #251

PR #178, opened 2023-09-07
doSave and SaveHelper done

rnaiser:


src/main/java/de/uni_tuebingen/ub/nppm/db/SaveHelper.java
```diff
@@ -43,8 +44,42 @@ public static void updateAttribute(String table, String attribute, String value,
         updateAttribute(table, attribute, value, Integer.parseInt(id));
     }
 
+    public static void updateAttribute(String table, String attribute, String value, String formularAttribut, int id, String cond) throws Exception {
+        try ( Session session = getSession()) {
+            session.getTransaction().begin();
+            NativeQuery query = session.createNativeQuery("UPDATE " + table + " SET " + attribute + "= :value WHERE " + formularAttribut + "= :id" + cond);
```
mtrojan-ub: cond also contains lots of values which must be escaped in a proper way

src/main/java/de/uni_tuebingen/ub/nppm/db/SaveHelper.java
```diff
@@ -43,8 +44,42 @@ public static void updateAttribute(String table, String attribute, String value,
...
+    public static void updateAttribute(String table, String attribute, String value, String formularAttribut, String id, String cond) throws Exception {
+        updateAttribute(table, attribute, value, formularAttribut, Integer.parseInt(id), cond);
+    }
+
+    public static void insertNewAttribute(String table, String attribute, String formularAttribut, String field, String value, int id, String Idvalue) throws Exception {
+        try ( Session session = getSession()) {
+            session.getTransaction().begin();
+
+            NativeQuery query = session.createNativeQuery("INSERT INTO " + table
+                    + " (" + attribute + ", " + formularAttribut + field + ")"
```
mtrojan-ub, line 92: All parts of the query must be escaped properly, not only id + value, but also attribute, formularAttribut and so on

src/main/webapp/dosave.jsp
```diff
@@ -5,368 +5,370 @@
...
-                                        " and UeberlieferungID in (select h_u.ID from handschrift_ueberlieferung h_u, quelle_inedition q_i where q_i.ID=" +
-                                        request.getParameter(datenfeld.toLowerCase() + "[" + i + "]_entryid") + " and q_i.QuelleID=h_u.QuelleID)"
-                                      );
+                                    if (editionLst.size() > 0) {
+                                        old_ed = editionLst.get(editionLst.size() - 1).get("EditionID").toString();
+                                    }
+                                }
+                                SaveHelper.insertOrUpdateSql(sql);
+                                if (!ed.equals("")) {
+                                    SaveHelper.insertOrUpdateSql("Update ueberlieferung_edition set EditionID='" + ed + "' where EditionID=" + old_ed
```
mtrojan-ub, line 317: In an optimal case, all queries should be built by the SaveHelper, so the JSP template should not build any queries by itself

rnaiser:
Administration-->Auswahlfelder-->bearbeiten, jetz kann man auch ' ' und " " benutzen (anlegen)

PR #181, opened 2023-09-18
moved Content to admin and secured Administrator login

rnaiser:
Inhalt ist nun in Admin bereich, ausserdem habe ich Adminstration abgesichert, davor konnte man wenn man angemeldet ist und kein admin ist in der url  eingeben /administration und man war drinne, jetzt fliegt man raus (wenn man hacken will) und ist im login bereich


rnaiser:
man konnte  wenn man ein neuen  Einzelbeleg anlegt und die checkbox  Datierung ungewiss anclickte , speicherte es nicht, nun tut es

src/main/java/de/uni_tuebingen/ub/nppm/db/SaveHelper.java
```diff
@@ -47,6 +50,57 @@ public static List<Map<String, String>> selectAttribute(String table, String att
         return results;
     }
 
+    public static List<Map> selectAttributeMap(String table, String attributeString, int id) throws Exception {
+        List<Map> results = new ArrayList<>();
+
+        try ( Session session = getSession()) {
+            session.getTransaction().begin();
```
mtrojan-ub: If you only select something from the database without modifying, there is no transaction needed

src/main/java/de/uni_tuebingen/ub/nppm/db/SaveHelper.java
```diff
@@ -47,6 +50,57 @@ public static List<Map<String, String>> selectAttribute(String table, String att
         return results;
     }
 
+    public static List<Map> selectAttributeMap(String table, String attributeString, int id) throws Exception {
```
mtrojan-ub: All function names that only use a SELECT statement should usually start with get... instead of select...

src/main/java/de/uni_tuebingen/ub/nppm/db/SaveHelper.java
```diff
@@ -47,6 +50,57 @@ public static List<Map<String, String>> selectAttribute(String table, String att
...
+        List<Map> results = new ArrayList<>();
+
+        try ( Session session = getSession()) {
+            session.getTransaction().begin();
+            String sql = "SELECT " + attributeString + " FROM " + table + " WHERE ID = " + id; // SQL-Abfrage erstellen
+
+            NativeQuery query = session.createNativeQuery(sql);
```
mtrojan-ub: If you want to execute a plain SQL statement & get the results as `List<Object[]>,` you can use the parent's function `AbstractBase.getListNative(String sql)`. Then you have to write less code, you don't even have open a session manually for this.

src/main/java/de/uni_tuebingen/ub/nppm/db/SaveHelper.java
```diff
@@ -47,6 +50,57 @@ public static List<Map<String, String>> selectAttribute(String table, String att
...
+
+    public static List<Map<String, String>> getMapField(String zieltabelle, int id) throws Exception {
+        List<Map<String, String>> results = new ArrayList<>();
+
+        try ( Session session = getSession()) {
+            session.getTransaction().begin();
+            String sql = "SELECT * FROM " + zieltabelle + " WHERE ID = " + id; // SQL-Abfrage erstellen
+
+            SQLQuery query = session.createSQLQuery(sql);
+            query.setResultTransformer(Criteria.ALIAS_TO_ENTITY_MAP);
```
mtrojan-ub: Same here. You should not use the ResultTransformer directly since it is deprecated and we have to replace it later when migrating to Hibernate 6. Instead, just use `AbstractBase.getMappedList(String query)`, then you also have to write less code again.

src/main/java/de/uni_tuebingen/ub/nppm/db/SaveHelper.java
```diff
@@ -47,6 +50,57 @@ public static List<Map<String, String>> selectAttribute(String table, String att
...
+
+            List<Map<String, String>> queryResults = query.list();
+
+            for (Map<String, String> rowMap : queryResults) {
+                results.add(rowMap);
```
mtrojan-ub: If you just want to get a result & transform it in a certain way, maybe we should even move these functions to AbstractBase? But i am not sure what the benefit for you is to get this kind of result datatype, so maybe we should discuss this.

src/main/java/de/uni_tuebingen/ub/nppm/db/SaveHelper.java
```diff
@@ -67,7 +121,7 @@ public static void updateAttribute(String table, String attribute, String value,
         updateAttribute(table, attribute, value, Integer.parseInt(id));
     }
 
-     public static void updateAttribute(String table, String attribute, String value, Map<String, String> andConditions) throws Exception {
+    public static void updateAttribute(String table, String attribute, String value, Map<String, String> andConditions) throws Exception {
```
mtrojan-ub: By the way, it seems like the word `attribute` is overused a lot. In my understanding, you want to update one or more existing rows and set a column to a specific value using certain conditions. So  a properly named function should be something like this:

`public static void update(String table, String column, String value, Map<String,String> andConditions) throws Exception`

Maybe it would be efficient to overload this and create main implementation:
`public static void update(String table, Map<String,String> columnsAndValues, Map<String,String> andConditions)`

And then changing the existing function just calling the other one and creating a map out of the single given column + value.

src/main/java/de/uni_tuebingen/ub/nppm/db/SaveHelper.java
```diff
@@ -82,7 +136,22 @@ public static void updateAttribute(String table, String attribute, String value,
         }
     }
 
+    public static void insertNewAttribute(String table, String attributeOne, String attributeTwo, String attributeThree, int valueOne, int valueTwo, String valueThree) throws Exception {
```
mtrojan-ub: For insert, i would recommend the same things as for the update.

It doesn't make much sense to hardcode an inserting function with a fixed number of parameters.
Instead, the function header should look like this:

`public static void insert(String table, Map<String,String> columnsAndValues) throws Exception`

And by the way: If these update/insert implementations are good enough, we could even think about moving them to the AbstractBase later (but maybe we should rather keep them protected then).

src/main/java/de/uni_tuebingen/ub/nppm/servlet/backend/ContentServlet.java
```diff
@@ -31,8 +30,18 @@ protected void generatePage(HttpServletRequest request, HttpServletResponse resp
             RequestDispatcher rd = request.getRequestDispatcher("tinyMce.jsp");
             rd.include(request, response);
         } else {
-        //show fileManagement
-
+            //show fileManagement
+            out.println("<div id=\"titel\">");
```
mtrojan-ub: Can this HTML header code be moved from the Servlet to the JSP template?

src/main/webapp/administration.jsp
```diff
@@ -7,7 +8,13 @@
 <%@ include file="functions.jsp" %>
 
 
-<%    Language.setLanguage(request);
+<%
+    int benutzerID = ((Integer) session.getAttribute("BenutzerID")).intValue();
+    Benutzer benutzer = BenutzerDB.getById(benutzerID);
+    boolean isAdmin = benutzer.isAdmin();
+    if (isAdmin) {
```
mtrojan-ub: This change should be applied to multiple admin pages. In fact, it would be more efficient to handle this in the AbstractBackendServlet. There is already a function called `isLoginRequired()` that can be overridden by inherited classes. I think it would be a good idea to add a similar function there which is called `isAdminRequired()` that can be overridden by all Backend Servlet implementations. Then we can handle it in a similar way inside the initRequest function, and it should be the least overall effort.

src/main/webapp/dosave.jsp
```diff
@@ -149,19 +150,34 @@
...
                         for (int i = 0; i < zielattributArray.length; i++) {
-                            if (request.getParameter(combinedFeldnamenArray[i]) != null && request.getParameter("Genauigkeit" + combinedFeldnamenArray[i]) != null) {
-                                if (i > 0) {
-                                    sql += ", ";
-                                }
-                                sql += zielattributArray[i] + "=\"" + ((!request.getParameter(combinedFeldnamenArray[i]).equals("")) ? request.getParameter(combinedFeldnamenArray[i]) : "0") + "\", ";
-                                sql += "Genauigkeit" + zielattributArray[i] + "=\"" + ((!request.getParameter("Genauigkeit" + combinedFeldnamenArray[i]).equals("")) ? request.getParameter("Genauigkeit" + combinedFeldnamenArray[i]) : "0") + "\"";
-                                sqlValid = true;
+                            String fieldValue = request.getParameter(combinedFeldnamenArray[i]);
+                            String genauigkeitValue = request.getParameter("Genauigkeit" + combinedFeldnamenArray[i]);
```
mtrojan-ub, line 166: As we discussed it yesterday, i think you were not yet able to test this in detail. Have you tried using this e.g. with certain date fields?

src/main/webapp/fileManagement.jsp
```diff
@@ -13,189 +13,196 @@
...
-        <%
+<%@ include file="configuration.jsp" %>
+
+
+    <%
+        int benutzerID = ((Integer) session.getAttribute("BenutzerID")).intValue();
+        Benutzer benutzer = BenutzerDB.getById(benutzerID);
+        boolean isAdmin = benutzer.isAdmin();
+
+        if (isAdmin) {
```
mtrojan-ub: Same here, this should be checked in the servlet.

src/main/webapp/tinyMce.jsp
```diff
@@ -10,81 +10,84 @@
 <%@ include file="functions.jsp" %>
 <%@ page import="de.uni_tuebingen.ub.nppm.util.*" isThreadSafe="false" %>
 
-<%    if (AuthHelper.isAdminLogin(request)) {
 
+
+<%
+    int benutzerID = ((Integer) session.getAttribute("BenutzerID")).intValue();
+    Benutzer benutzer = BenutzerDB.getById(benutzerID);
+    boolean isAdmin = benutzer.isAdmin();
```
mtrojan-ub: And same here again

mtrojan-ub (changes requested):
Thanks, @rnaiser! I had a more intense look at your changes and left a lot of comments, please have a look.

src/main/java/de/uni_tuebingen/ub/nppm/db/SaveHelper.java
rnaiser: but the Abstract method have no return type of List<Map<String, String>>

rnaiser: hab eine lösung gefunden, siehe unten : created new function in AbstractBase

src/main/java/de/uni_tuebingen/ub/nppm/db/SaveHelper.java
rnaiser: done:  look at commit: improvement of select functions

src/main/java/de/uni_tuebingen/ub/nppm/db/SaveHelper.java
rnaiser: done:  look at commit: improvement of select functions

src/main/java/de/uni_tuebingen/ub/nppm/db/SaveHelper.java
rnaiser: done:  look at commit: use of AbstractBase.getListNative

src/main/java/de/uni_tuebingen/ub/nppm/db/SaveHelper.java
rnaiser: i moved it to AbstractBase look at created new function in AbstractBase,
i get this kind of result datatype cause i changed in dosave : from : List<Map> attributes = SaveHelper.getMappedList("SELECT " + zielAttribut + " FROM " + zieltabelle + " WHERE " + formularAttribut + "='" + id + "'" + cond);
to:    List<Map<String, String>> attributes = SaveHelper.getAttribute(zieltabelle, zielAttribut, condMap);  it functions perfect of course i could change it maybe back to jsut List<Map> attributes = .... but than i think why ... when it is now like this and maybe it was only working this way .... if you want me to change or try it with just List<Map> attributes then let me know

src/main/java/de/uni_tuebingen/ub/nppm/db/SaveHelper.java
rnaiser: done! look at commit: update and getAttribute Method
Instead of 3 methods, I now have one and adapted it in the dosave. I also had to expand the public static List<Map<String, String>> getAttribute(String table, String attribute, Map<String, String> andConditions) because otherwise it would be in dosave (combined) wouldn't work otherwise.

src/main/java/de/uni_tuebingen/ub/nppm/servlet/backend/ContentServlet.java
rnaiser: Nein das kann es nicht, da wenn ich den header hier habe kann ich die informationen Geben wei Datei & Bilder Verwalten 
 Wenn sie eine Datei löschen, dann ist auch jede Verknüpfung im Programm gelöscht, auch wenn Sie      		
  auch sonst bekomme ich probleme wenn eine Datei hochgeladen wird, das alte informationen nicht mehr angezeigt werden wie Datei kann nicht gelöscht werden, ich habe es wirklich ausgetestet, hier sollte das so beiben.

src/main/java/de/uni_tuebingen/ub/nppm/db/SaveHelper.java
rnaiser: done look at comment: insert function in saveHelper renewed

src/main/webapp/administration.jsp
rnaiser: done look at commit: isAdminRequired function

src/main/webapp/fileManagement.jsp
rnaiser: done look at commit: isAdminRequired function

src/main/webapp/tinyMce.jsp
rnaiser: done look at commit: isAdminRequired function

rnaiser:
admin.benutzer.neu.jsp: benutzer.setActiv(true) hinzugefügt so war es auch im orginal

einstellungen.speichern.jsp:  benutzer.setAktiv(request.getParameter("Aktiv")!=null && request.getParameter("Aktiv").equals("on"));  zu benutzer.setAdmin("on".equals(request.getParameter("Administrator")));  sonst Null pointer exception

LoginServlet.java:  if (benutzer == null || !benutzer.isAktiv()) {
        response.sendError(HttpServletResponse.SC_UNAUTHORIZED, "Zugriff nicht erlaubt, Ihr Administrator muss Sie auf aktiv schalten !");
        return;
    }  falls Benutzer nicht Activ ist

rnaiser:
Bugfix: Bei Textzeugen wenn keine Datierung angegeben wird soll (auch im orginal) 0(0.Jhd)-0(0.Jhd) angezeigt wird das tut es nun, davor kamm der error: 
processing [/forms/array.combined.jsp] at line [476]

473:                                 + i
474:                                 + "]\">"
475:                                 + row.get(combinedFeldnamen[j]
476:                                         + "VonJahr").toString()
477:                                 + "("
478:                                 + row.get(combinedFeldnamen[j]
479:                                         + "VonJahrhundert").toString()


Stacktrace:] with root cause
	java.lang.NullPointerException

src/main/java/de/uni_tuebingen/ub/nppm/db/DeleteDB.java
```diff
@@ -0,0 +1,35 @@
+package de.uni_tuebingen.ub.nppm.db;
+
+import static de.uni_tuebingen.ub.nppm.db.AbstractBase.getSession;
+import org.hibernate.Session;
+import org.hibernate.query.NativeQuery;
+
+public class DeleteDB extends AbstractBase {
```
mtrojan-ub: Introducing a new class named `DeleteDB` and adding methods in it that do select & update-Operations doesn't really make sense. SELECT & UPDATE has nothing to do with delete. 

Maybe we just move these functions to e.g. DatenbankDB?

src/main/java/de/uni_tuebingen/ub/nppm/db/DeleteDB.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public static void updateAttribute(String table, String attribute, int id) throws Exception {
+        try ( Session session = getSession()) {
+            session.getTransaction().begin();
+            NativeQuery query = session.createNativeQuery("UPDATE " + table + " SET " + attribute + "= :value WHERE ID= :id");
+            query.setParameter("value", null);
```
mtrojan-ub: Seems like you just support value to be hardcoded null, which doesn't make much sense to me in a generic function. You should add a `value` parameter after `attribute` that can be null, and let the caller decide.

mtrojan-ub: And in fact i think you should change the function naming & signature

I also think we already talked about this 1-2 weeks before. The main implementation should be in an additional function:

`public static void update(String table, int id, Map<String,String> columnsAndValues);`

And this function should only redirect with an implementation similar to this one:

```
public static void update(String table, int id, String column, String value) {
    update(table, id, Map.ofEntries(entry(column, value)));
}
```

src/main/java/de/uni_tuebingen/ub/nppm/db/DeleteDB.java
```diff
@@ -0,0 +1,35 @@
...
+            query.setParameter("id", id);
+            query.executeUpdate();
+            session.getTransaction().commit();
+        }
+    }
+
+    public static String selectAttribute(String table, String attribute, int id) throws Exception {
```
mtrojan-ub: We should change the select... function into 2 functions that act in a similar way (see my comments related to the update... function).

```
public static Map<String,String> select(String table, int id, List<String> columns) {
    // ... slightly adjust previous implementation here
}

public static String select(String table, int id, String column) {
    Map<String, String> columnsAndValues = select(table, id, Arrays.asList(column));
    return columnsAndValues[column];
}
```

src/main/java/de/uni_tuebingen/ub/nppm/db/SaveHelper.java
```diff
@@ -22,52 +25,72 @@ public static void insertMetaData(String form, int id, int benutzerID, int grupp
         insertOrUpdate(sql);
     }
 
-    public static List<Map<String, String>> selectAttribute(String table, String attribute, Map<String, String> andConditions) throws Exception {
+    public static List<Map<String, String>> getAttribute(String table, String attribute, Map<String, String> andConditions) throws Exception {
```
mtrojan-ub: All these get... functions should also be overloaded select... functions as specified in the previous comments. 

I mean, in the end i'm not 100% sure whether the functions should all be called select... or get... but we should name all functions the same that also do similar things and just overload them, instead of having lots of different function names for functions that do basically the same.

mtrojan-ub: Another comment related to the signature: If we only select a single attribute, i don't think it makes much sense to use `List<Map<String,String>>` as a result type. A simple `List<String>` should be enough. I think  a `List<Map<String,String>>` only makes sense if we select multiple attributes at the same time.

src/main/java/de/uni_tuebingen/ub/nppm/db/SaveHelper.java
```diff
@@ -22,52 +25,72 @@ public static void insertMetaData(String form, int id, int benutzerID, int grupp
...
+
+                results.add(rowMap);
+            }
         }
+        return results;
     }
 
-    public static void updateAttribute(String table, String attribute, String value, String id) throws Exception {
-        updateAttribute(table, attribute, value, Integer.parseInt(id));
+    public static List<Map<String, String>> getMapField(String zieltabelle, int id) throws Exception {
```
mtrojan-ub: If this is a simple variant of an AbstractBase function, it should be moved into that class.

src/main/java/de/uni_tuebingen/ub/nppm/servlet/LoginServlet.java
```diff
@@ -15,50 +15,59 @@
...
-        }
+    if (login == null || login.isEmpty() || !BenutzerDB.hasLogin(login)) {
+        response.sendError(HttpServletResponse.SC_UNAUTHORIZED, "Benutzer existiert nicht");
+        return;
+    }
+
+    Benutzer benutzer = BenutzerDB.getByLogin(login);
+
+    if (benutzer == null || !benutzer.isAktiv()) {
+        response.sendError(HttpServletResponse.SC_UNAUTHORIZED, "Zugriff nicht erlaubt, Ihr Administrator muss Sie auf aktiv schalten !");
```
mtrojan-ub: There should be different messages for when the user is null or when he is not active.

src/main/java/de/uni_tuebingen/ub/nppm/servlet/backend/AbstractBackendServlet.java
```diff
@@ -26,6 +27,19 @@ protected boolean isLoginRequired() {
...
     }
 
+    protected boolean isAdminRequired(HttpServletRequest request, HttpServletResponse response) throws Exception {
+    Integer benutzerID = (Integer) request.getSession().getAttribute("BenutzerID");
+
+    if (benutzerID != null) {
+        Benutzer benutzer = BenutzerDB.getById(benutzerID);
+
+        if (benutzer != null) {
+            return benutzer.isAdmin();
```
mtrojan-ub: - The implementation of isAdminRequired in this class should only return a default value. All inheriting servlets that differ from this default value must override the function.
- The check if the current iser is an admin or not should be moved to the initRequest method, similar to the isLoginRequired mechanism.

src/main/java/de/uni_tuebingen/ub/nppm/servlet/backend/AdministrationAuswahlServlet.java
```diff
@@ -4,16 +4,23 @@
...
     @Override
     protected String getTitle() {
         return "administration";
     }
 
     @Override
     protected void generatePage(HttpServletRequest request, HttpServletResponse response) throws Exception {
-        RequestDispatcher rd = request.getRequestDispatcher("admin.auswahlfelder.jsp");
-        rd.include(request, response);
+          //Use isAdminRequired to check if the user is an administrator.
```
mtrojan-ub: This check will be done by the parent function. The generatePage function should stay as before, you should only override the isAdminRequired function.

src/main/java/de/uni_tuebingen/ub/nppm/servlet/backend/AdministrationBaumstrukturServlet.java
```diff
@@ -18,10 +18,18 @@ protected String getTitle() {
...
     protected void generatePage(HttpServletRequest request, HttpServletResponse response) throws Exception {
-        RequestDispatcher rd = request.getRequestDispatcher("admin.baumstruktur.jsp");
-        rd.include(request, response);
+         //Use isAdminRequired to check if the user is an administrator.
+        if (isAdminRequired(request, response)) {
+            RequestDispatcher rd = request.getRequestDispatcher("admin.baumstruktur.jsp");
+            rd.include(request, response);
+        }else {
+            //If the user is not an administrator, redirect them to logout.jsp.
+            RequestDispatcher dispatcher = request.getRequestDispatcher("/logout.jsp");
```
mtrojan-ub: This check will be done by the parent function. The generatePage function should stay as before, you should only override the isAdminRequired function.

src/main/java/de/uni_tuebingen/ub/nppm/servlet/backend/AdministrationServlet.java
```diff
@@ -5,15 +5,22 @@
...
     protected void generatePage(HttpServletRequest request, HttpServletResponse response) throws Exception {
-        RequestDispatcher rd = request.getRequestDispatcher("administration.jsp");
-        rd.include(request, response);
+        //Use isAdminRequired to check if the user is an administrator.
+        if (isAdminRequired(request, response)) {
+            RequestDispatcher rd = request.getRequestDispatcher("administration.jsp");
+            rd.include(request, response);
+        } else {
+            //If the user is not an administrator, redirect them to logout.jsp.
+            RequestDispatcher dispatcher = request.getRequestDispatcher("/logout.jsp");
```
mtrojan-ub: This check will be done by the parent function. The generatePage function should stay as before, you should only override the isAdminRequired function.

src/main/java/de/uni_tuebingen/ub/nppm/servlet/backend/ContentServlet.java
```diff
@@ -26,46 +25,62 @@ public class ContentServlet extends AbstractBackendServlet {
     protected void generatePage(HttpServletRequest request, HttpServletResponse response) throws Exception {
         PrintWriter out = response.getWriter();
 
-        // show TinyMCE
-        if (request.getParameter("loadFile") != null) {
-            RequestDispatcher rd = request.getRequestDispatcher("tinyMce.jsp");
-            rd.include(request, response);
+        if (!isAdminRequired(request, response)) {
+            //If the user is not an administrator, redirect them to logout.jsp.
+            RequestDispatcher dispatcher = request.getRequestDispatcher("/logout.jsp");
```
mtrojan-ub: This check will be done by the parent function. You should only override the isAdminRequired function.

src/main/webapp/einstellungen.speichern.jsp
```diff
@@ -69,7 +69,7 @@
       benutzer.setLogin(request.getParameter("Benutzername"));
       benutzer.setVorname(request.getParameter("Vorname"));
       benutzer.setNachname(request.getParameter("Nachname"));
-      benutzer.setAdmin(request.getParameter("Administrator").equals("on"));
+      benutzer.setAdmin("on".equals(request.getParameter("Administrator")));
```
mtrojan-ub, line 72: Does changing this line really make a difference?

rnaiser: yes it does !!!  this way it does not get a Null Pointer Exception if request.getParamerter("Administrator") is null and than you make .equalts("on") gives you a Exception the otherway around not

mtrojan-ub: Interesting explanation, thanks!

rnaiser:
should i make this :   protected boolean isAdminRequired(HttpServletRequest request, HttpServletResponse response) throws Exception { ... }   it to a void function ?

src/main/java/de/uni_tuebingen/ub/nppm/servlet/backend/AdministrationServlet.java
rnaiser: done: look at commit: isAdminRequierd function and AdministratorServlet

mtrojan-ub:
No, isAdminRequired must always return a boolean. As discussed, i would have also preferred a class variable but since the purpose is to override this in the child classes, this can only be achieved in java using a function instead of a class member variable, similar to isLoginRequired.

src/main/java/de/uni_tuebingen/ub/nppm/servlet/backend/AdministrationAuswahlServlet.java
rnaiser: done: look at commit: Administration

src/main/java/de/uni_tuebingen/ub/nppm/servlet/backend/AdministrationBaumstrukturServlet.java
rnaiser: done: look at commit: Administration

src/main/java/de/uni_tuebingen/ub/nppm/servlet/backend/AdministrationServlet.java
rnaiser: done: look at commit: Administration

src/main/java/de/uni_tuebingen/ub/nppm/servlet/backend/ContentServlet.java
rnaiser: done: look at commit: Administration

src/main/java/de/uni_tuebingen/ub/nppm/servlet/LoginServlet.java
rnaiser: done:  look at commit : Login Servlet message

src/main/java/de/uni_tuebingen/ub/nppm/servlet/backend/AbstractBackendServlet.java
rnaiser: done look at commit: Administration

src/main/java/de/uni_tuebingen/ub/nppm/db/SaveHelper.java
rnaiser: done: look at commit: changed Map<String, String> to Map

i can not use List<String> cause in do save i work with List<Map> attributes before change it was List<Map<String, String> attributes and i changed it back to List<Map> like it was original. and changed the function name now to public static List<Map> getAttribute(String table, String attribute, Map andConditions) throws Exception

rnaiser: Me and Julian have tested it, the have diverent return types like List<Map>, String and also:

//Diese funktion ist notwendig für ist<Map> attributes = SaveHelper.getAttributeMap(zieltabelle, zielAttributStr, id); in doSave, da sie ein Single-Objekt zurück gibt
  //  und sonst ein Exception auftritt wenn ich die public static List<Map> getAttribute(String table, String attribute, Map andConditions) throws Exception { benützen würde
    public static List<Map> getAttributeMap(String table, String attributeString, int id) throws Exception {

src/main/java/de/uni_tuebingen/ub/nppm/db/DeleteDB.java
rnaiser: Done look at commit: problem DeleteDB solved   
DeleteDB wurde gelöscht, habe die SaveHelper funktionen kann man hier verwenden und die habe ich dann in die Abstract Base verschoben

src/main/java/de/uni_tuebingen/ub/nppm/db/DeleteDB.java
rnaiser: Done look at commit: problem DeleteDB solved

src/main/java/de/uni_tuebingen/ub/nppm/db/DeleteDB.java
rnaiser: Done look at commit: problem DeleteDB solved

rnaiser:
i have done everything now, you can merge it !

PR #183, opened 2023-10-06
Extend NEG for Migration from DMP

juro-ub:
Workshop 26.09.
Fields are missing in NeG:

1. pal_abgrenzung
2. nr_in_strukt
3. inh_abgrenzung

4. seite
5. raster
6. schreiber

This PR includes the implementation of the 6 Fields (Database, Frontend, Backend)

### Backend:
image

image

### Frontend
image

### DIFF
--- a/src/main/webapp/gast/einzelbeleg.jsp
+++ b/src/main/webapp/gast/einzelbeleg.jsp
@@ -166,6 +166,39 @@
         </jsp:include>
       </td>
     </tr>
+    <tr>
+      <th><% Language.printDatafield(out, session, "einzelbeleg", "PalAbgrenzung"); %></th>
+      <td>
+        <jsp:include page="../inc.erzeugeFormular.jsp">
+          <jsp:param name="ID" value="<%= id %>"/>
+          <jsp:param name="Formular" value="einzelbeleg"/>
+          <jsp:param name="Datenfeld" value="PalAbgrenzung"/>
+          <jsp:param name="Readonly" value="yes"/>
+        </jsp:include>
+      </td>
+    </tr>
+    <tr>
+      <th><% Language.printDatafield(out, session, "einzelbeleg", "InhAbgrenzung"); %></th>
+      <td>
+        <jsp:include page="../inc.erzeugeFormular.jsp">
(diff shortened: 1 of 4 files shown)


mtrojan-ub:
As discussed in this week's workshop, it would make much sense to also add the other fields here:
- seite
- raster
- schreiber

Some more thoughts:
- We did not discuss yet in which form they should be presented in the frontend (you added them to the search result),  maybe we should  ask @ThoFischer first whether that's OK
- Can you please add a full screenshot from the non-compact backend page where all these are shown? Since all of them belong to a similar topic, maybe we should put all of them together into a blue box (similar to "Datierung") and find a proper group name?

juro-ub:
> As discussed in this week's workshop, it would make much sense to also add the other fields here:
> 
> * seite
> * raster
> * schreiber
> 
> Some more thoughts:
> 
> * We did not discuss yet in which form they should be presented in the frontend (you added them to the search result),  maybe we should  ask @ThoFischer first whether that's OK
> * Can you please add a full screenshot from the non-compact backend page where all these are shown? Since all of them belong to a similar topic, maybe we should put all of them together into a blue box (similar to "Datierung") and find a proper group name?

I have added the 3 fields to a group named Raster

image

src/main/webapp/gast/einzelbeleg.jsp
mtrojan-ub, line 1: Merge conflict: Please rename 27.sql to 29.sql

updates/27.sql
mtrojan-ub, line 1: Merge conflict: Please rename 27.sql to 29.sql

mtrojan-ub:
Thanks for adding the box!

- As far as i understand it, the other new fields 1-3 belong to a similar topic. Maybe we should also move them into the new blue box?
- Maybe we should also find a better title for the box?

=> i will ask @ThoFischer via email & put you on CC

updates/27.sql
juro-ub: conflict is solved

mtrojan-ub:
First i thought that 10 might be a bit short as field length for all of these fields, but it looks like all current Values in DMP seem to be only numbers or abbreviations, so i guess it's OK like this.

Thanks! I will merge it!

PR #184, opened 2023-10-12
Selektion Kontext

juro-ub:
New Field "Kontext"

image

### DIFF
--- a/src/main/webapp/einzelbeleg.jsp
+++ b/src/main/webapp/einzelbeleg.jsp
@@ -120,6 +120,16 @@
 								<jsp:param name="rows" value="5" />
 							</jsp:include></td>
 					</tr>
+                                        <tr>
+                                            <td width="200" valign="top">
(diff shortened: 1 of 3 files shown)


mtrojan-ub:
Is this basically the same as #149?

juro-ub:
> Is this basically the same as #149?

Yes it is basically the same. This PR will add a Select Field for Kontext and #149 will add a Select Field for "Textkritische Anmerkung"

mtrojan-ub:
Can you please solve the open conflict regarding `updates/24.sql` and update the screenshot so we can see where exactly the field will be? /The new additional field from #149 is not visible yet in the screenshot.)

juro-ub:
> Can you please solve the open conflict regarding `updates/24.sql` and update the screenshot so we can see where exactly the field will be? /The new additional field from #149 is not visible yet in the screenshot.)

Conflicts resolved and screenshot updated

PR #185, opened 2023-10-17
Fix for Previous PR #149 

juro-ub:
Add FK

### DIFF
--- a/updates/23.sql
+++ b/updates/23.sql
@@ -17,4 +17,4 @@ VALUES ("einzelbeleg", "KritikSelektion", "Textkritische Anmerkung", "select", 0
 
 ALTER TABLE einzelbeleg ADD KritikID INT DEFAULT -1;
 
-ALTER TABLE einzelbeleg ADD CONSTRAINT FK_EinzelbelegKritik_1 FOREIGN KEY (KritikID) REFERENCES selektion_kritik(ID);
+ALTER TABLE einzelbeleg ADD CONSTRAINT einzelbeleg_KritikID FOREIGN KEY (KritikID) REFERENCES selektion_kritik(ID);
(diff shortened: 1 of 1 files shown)


updates/23.sql
```diff
@@ -16,3 +16,5 @@ INSERT INTO datenbank_mapping (Formular, Datenfeld, de_Beschriftung, Feldtyp, Ar
 VALUES ("einzelbeleg", "KritikSelektion", "Textkritische Anmerkung", "select", 0, "einzelbeleg", "KritikID", "selektion_kritik", "einzelbeleg", "Text-critical note", "Note critique sur le texte", "Nota cum textu critico");
 
 ALTER TABLE einzelbeleg ADD KritikID INT DEFAULT -1;
+
+ALTER TABLE einzelbeleg ADD FOREIGN KEY (KritikID) REFERENCES selektion_kritik(ID);
```
mtrojan-ub: All keys should also have distinct names

juro-ub: i have added a distinct name

updates/23.sql
```diff
@@ -16,3 +16,5 @@ INSERT INTO datenbank_mapping (Formular, Datenfeld, de_Beschriftung, Feldtyp, Ar
 VALUES ("einzelbeleg", "KritikSelektion", "Textkritische Anmerkung", "select", 0, "einzelbeleg", "KritikID", "selektion_kritik", "einzelbeleg", "Text-critical note", "Note critique sur le texte", "Nota cum textu critico");
 
 ALTER TABLE einzelbeleg ADD KritikID INT DEFAULT -1;
+
+ALTER TABLE einzelbeleg ADD CONSTRAINT FK_EinzelbelegKritik_1 FOREIGN KEY (KritikID) REFERENCES selektion_kritik(ID);
```
mtrojan-ub: If you execute `SHOW CREATE TABLE einzelbeleg` you can see that all other existing foreign keys follow a naming scheme, e.g. `einzelbeleg_BearbeitungsstatusID`, `einzelbeleg_EditionID` and so on. Please change the name to `einzelbeleg_KritikID` according to the existing convention.

juro-ub: Fixed

PR #188, opened 2023-10-20
Fixes

juro-ub:
- Remove Obsolete Imports
- Make saveOrUpdate Function abstract
- Fix NullPointerException

### DIFF
--- a/updates/9.sql
+++ b/updates/9.sql
@@ -19,6 +19,10 @@
 #     - If they are in a n:m table, they should be NOT NULL without default
 #     - If they are in a column, they should be NOT NULL with default -1 or the corresponding default value matching "-" or "--" (NOT "?"!)
 #   - So far we will NOT set any key to use ON DELETE CASCADE or ON DELETE UPDATE. This might be done at a later point if needed.
+
+#Otherwise an error will appear. Will be restored at the end of the script
+SET FOREIGN_KEY_CHECKS=0;
+
 ALTER TABLE bemerkung MODIFY COLUMN ID INT UNSIGNED AUTO_INCREMENT;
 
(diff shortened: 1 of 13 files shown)


mtrojan-ub:
I'm not entirely sure whether the removal of the sql imports are really working on all subpages (e.g. whether inc.modul.php provides its own includes in the old state). Did you test it thoroughly?

updates/9.sql
```diff
@@ -19,6 +19,10 @@
 #     - If they are in a n:m table, they should be NOT NULL without default
 #     - If they are in a column, they should be NOT NULL with default -1 or the corresponding default value matching "-" or "--" (NOT "?"!)
 #   - So far we will NOT set any key to use ON DELETE CASCADE or ON DELETE UPDATE. This might be done at a later point if needed.
+
+#Otherwise an error will appear. Will be restored at the end of the script
+SET FOREIGN_KEY_CHECKS=0;
```
mtrojan-ub, line 24: There has been no error to me with the old migration. As we detected in a separate PR, simply using this might screw up foreign references. Did you test this thoroughly as well?

mtrojan-ub:
There's an open merge conflict regarding `updates/23.sql`

juro-ub:
> There's an open merge conflict regarding `updates/23.sql`

Conflict is solved now

updates/9.sql
juro-ub: Yes we have tested the script 9.sql. There are many statements that add FK Constraints.
Without the FOREIGN_KEY_CHECKS=0 statement the script runs in an error

mtrojan-ub: Can you please post an example for one of these errors?

PR #189, opened 2023-10-26
update funktions with two Maps

rnaiser:


rnaiser:
Bug war bei: einzelbeleg --> Tab/Reiter Textkrikit und wenn man dann neu anlegt und Bemerkung ausfüllt kommt ein error:
der sql string war nicht richtig z.b 
INSERT INTO einzelbeleg_textkritik SET EinzelbelegID = '1', EditionID = 'NULL', HandschriftID = 'NULL', Variante = '', Bemerkung = 'Peter Pan'   Broblem war das NULL in "" war nun ist das nicht mehr der fall: z.b INSERT INTO einzelbeleg_textkritik SET EinzelbelegID = '1', EditionID = '195', HandschriftID = NULL, Variante = 'ew4tr', Bemerkung = 'Halloween'

src/main/java/de/uni_tuebingen/ub/nppm/db/AbstractBase.java
```diff
@@ -260,10 +261,46 @@ protected static String buildAndConditions(Map<String, String> andConditions) {
...
+
+            if (andCondition.getKey().equals("QuelleBisJahrhundert") || andCondition.getKey().equals("QuelleVonJahrhundert") || andCondition.getKey().equals("VonJahrhundert") || andCondition.getKey().equals("BisJahrhundert") ) {
+                query.setParameter(andCondition.getKey(), andCondition.getValue(), StringType.INSTANCE);
+            } else {
+                query.setParameter(andCondition.getKey(), andCondition.getValue());
+            }
         }
     }
 
+     public static void updateMap(String table, Map<String, String> attributesAndValues, Map<String, String> andConditions) throws Exception {
```
mtrojan-ub: You can rename this function to "update" as well & simply use function overloading this way.

src/main/java/de/uni_tuebingen/ub/nppm/db/AbstractBase.java
```diff
@@ -260,10 +261,46 @@ protected static String buildAndConditions(Map<String, String> andConditions) {
     //Hilsfunktion zum setzen der Parameter
     protected static void registerAndConditions(Map<String, String> andConditions, NativeQuery query) {
         for (Map.Entry<String, String> andCondition : andConditions.entrySet()) {
-            query.setParameter(andCondition.getKey(), andCondition.getValue());
+
+            if (andCondition.getKey().equals("QuelleBisJahrhundert") || andCondition.getKey().equals("QuelleVonJahrhundert") || andCondition.getKey().equals("VonJahrhundert") || andCondition.getKey().equals("BisJahrhundert") ) {
```
mtrojan-ub: Why is this necessary? a generic function like this should not have special hardcoded handling for field names. If it is necessary to handle certain columns in a special way, this should be passed to the function as parameter as well.

src/main/java/de/uni_tuebingen/ub/nppm/db/AbstractBase.java
```diff
@@ -260,10 +261,46 @@ protected static String buildAndConditions(Map<String, String> andConditions) {
...
+                i++;
+
+            }
+
+            sql += buildAndConditions(andConditions);
+
+            NativeQuery query = session.createNativeQuery(sql);
+
+
+            registerAndConditions(attributesAndValues, query);
```
mtrojan-ub: attributesAndValues cannot be treated the same way as andConditions, so if you want to do it like this you will need 2 separate functions.

src/main/java/de/uni_tuebingen/ub/nppm/db/AbstractBase.java
rnaiser: done: look at commit: improvement of function update

src/main/java/de/uni_tuebingen/ub/nppm/db/AbstractBase.java
rnaiser: done: look at commit : improvement of function update

src/main/java/de/uni_tuebingen/ub/nppm/db/AbstractBase.java
rnaiser: done: look at commit: improvement of function update

PR #199, opened 2023-12-01
Stat

juro-ub:
### Stat page
image


mtrojan-ub:
Several points that differ from the notes in the ToDo-list:
- The entry point should be a link inside the view "Quellen". There should not be a menu entry for this.
- Due to your screenshot it looks like there is only a view for the list of the "Quellen" with all "Belege" per "Quelle". But there should be another view that shows all "Belege" across all "Quellen". 
- Also, since we are in the gast area, i think it is necessary to limit the statistics to only Quellen that are visible in public (zuVeroeffentlichen=1)

mtrojan-ub:
Seems like there is no process since january... please don't forget to update this PR + have a look at the comments soon, since we will need to merge this with the new design before the web relaunch.

juro-ub:
> * Beleg



> Several points that differ from the notes in the ToDo-list:
> 
> * The entry point should be a link inside the view "Quellen". There should not be a menu entry for this.
> * Due to your screenshot it looks like there is only a view for the list of the "Quellen" with all "Belege" per "Quelle". But there should be another view that shows all "Belege" across all "Quellen".
> * Also, since we are in the gast area, i think it is necessary to limit the statistics to only Quellen that are visible in public (zuVeroeffentlichen=1)

I have implemented the changes

src/main/webapp/gast/stat.jsp
```diff
@@ -0,0 +1,155 @@
...
+        </thead>
+       <tbody>
+            <%
+            for (Quelle q : lst) {
+                out.print("<tr onclick=\"window.location='/neg/gast/quelle?ID="+String.valueOf(q.getId())+"';\" style=\"cursor: pointer;\">");
+                out.print("<td>");
+                out.print(Utils.escapeHTML(q.getBezeichnung()));
+                out.print("</td>");
+                out.print("<td>");
+                out.print(EinzelbelegDB.countEinzelbelegByQuelleId(q.getId()));
```
mtrojan-ub: This looks inefficient => if 1 page in the view contains 100 entries, then there will be 100 extra SQL queries executed to render the view. You already have Quelle q in the model class => just add a function getEinzelbelege() to the model class with a OneToMany relation. Then you can use q.getEinzelbelege().size() instead.

src/main/webapp/gast/stat.jsp
```diff
@@ -0,0 +1,155 @@
...
+                %>
+            </th>
+            <th>
+                <b>Anzahl Belege</b>
+            </th>
+        </thead>
+       <tbody>
+            <%
+            for (Quelle q : lst) {
+                out.print("<tr onclick=\"window.location='/neg/gast/quelle?ID="+String.valueOf(q.getId())+"';\" style=\"cursor: pointer;\">");
```
mtrojan-ub: The full path including /neg should never be hardcoded because this wont work on the servers. Please build the URL using the Utils.getBaseUrl() prefix instead.

mtrojan-ub: Also, instead of the onclick event please use a proper hyperlink <a>...</a> instead inside the first <td> of your table. This will not only mark it in a way that is visible to the user, but also prevent CSP (Content Security Policy) issues. The onclick attribute should never be used in any context.

src/main/java/de/uni_tuebingen/ub/nppm/db/EinzelbelegDB.java
```diff
@@ -42,6 +42,18 @@ public static Einzelbeleg getFirstPublicEinzelbeleg() throws Exception {
     public static void insertBySql(String sql) throws Exception {
         insertOrUpdate(sql);
     }
+    
+    public static int countEinzelbelegByQuelleId(int qId) throws Exception {
```
mtrojan-ub: As mentioned, this function should be removed & implemented within the Quelle model class.

src/main/webapp/gast/stat.jsp
```diff
@@ -0,0 +1,155 @@
...
+                    <input name="sort" type="hidden" value="<%=sort %>"/>
+                    <input type="submit" />
+                </form>
+                <%
+                    out.println(html_sort_title_up(currentPage, filterTitle, recordsPerPage,jumpToID));
+                    out.println(html_sort_title_down(currentPage, filterTitle, recordsPerPage,jumpToID));
+                %>
+            </th>
+            <th>
+                <b>Anzahl Belege</b>
```
mtrojan-ub, line 141: We should also be able to sort by count belege up/down (asc/desc)

src/main/java/de/uni_tuebingen/ub/nppm/servlet/gast/StatistikServlet.java
```diff
@@ -0,0 +1,43 @@
...
+        return "stat";
+    }
+
+    @Override
+    protected String getTitle() {
+        return "Statistik";
+    }
+
+    @Override
+    protected void generatePage(HttpServletRequest request, HttpServletResponse response) throws Exception {
```
mtrojan-ub: I don't think we will need this in any other case except Quelle. Can we integrate this into the existing quelle servlet somehow? Otherwise we might rename it to QuelleStatistikServlet

src/main/webapp/gast/stat_overview.jsp
```diff
@@ -0,0 +1,9 @@
+<p>
+    <h2>Verfügbare Statistiken</h2>
```
mtrojan-ub: I also think we do not need this intermediate page. The link from "Quellen" should point directly to the anzahl_belege statistics.

src/main/webapp/gast/stat.jsp
```diff
@@ -0,0 +1,155 @@
...
+    if (request.getParameter("sort") != null) {
+       sort = request.getParameter("sort");
+    }
+    /*Set jumpToID for Quelle*/
+    String jumpToID = "";
+    if (request.getParameter("jumpToID") != null) {
+       jumpToID = request.getParameter("jumpToID");
+    }
+    /*
+    Calcualtion for pagination
```
mtrojan-ub: typo => Calculation

src/main/webapp/gast/stat.jsp
```diff
@@ -0,0 +1,155 @@
...
+
+    public String html_sort_title_up(int page, String filterTitle, int recordsPerPage, String jumpToID) {
+        return "<a class=\"sort-link\"href=\"?jumpToID="+jumpToID+"&sort=titleUp&page=anzahl_belege&filterTitle=" + filterTitle + "&recordsPerPage=" + recordsPerPage + "&currentPage=" + page + "\">sort up</a>";
+    }
+
+    public String html_sort_title_down(int page, String filterTitle, int recordsPerPage, String jumpToID) {
+        return "<a class=\"sort-link\"href=\"?jumpToID="+jumpToID+"&sort=titleDown&page=anzahl_belege&filterTitle=" + filterTitle + "&recordsPerPage=" + recordsPerPage + "&currentPage=" + page + "\">sort down</a>";
+    }
+
+    public void print_pagination(JspWriter out, int currentPage, int recordsPerPage, String filterTitle, int nOfPages, String sort, String jumpToID) throws Exception {
```
mtrojan-ub, line 88: Instead of the custom pagination elements => can we re-use the pagination that is also used in other templates?

Since this is a list, i understand that including the regular title.inc.jsp might not make sense, but maybe you can include forms/link.jsp in a similar way?

src/main/webapp/gast/stat.jsp
juro-ub: fixed!

src/main/java/de/uni_tuebingen/ub/nppm/db/EinzelbelegDB.java
juro-ub: i have removed the function

src/main/webapp/gast/stat.jsp
juro-ub: fixed!

src/main/java/de/uni_tuebingen/ub/nppm/servlet/gast/StatistikServlet.java
juro-ub: Yes i have included it in the Quellen Servlet

src/main/webapp/gast/stat.jsp
juro-ub: fixed

src/main/webapp/gast/stat_overview.jsp
juro-ub: I have removed the page

src/main/webapp/gast/stat.jsp
juro-ub: now the sort of belege is also possible

src/main/webapp/gast/stat.jsp
juro-ub: I think it is hard to reuse forms/link.jsp. 
link.jsp uses sql queries to determine the ID of the next and previous dataset,
while html_prev_button and html_next_button is based on pagination which calculate the next and previous page, based on the current page.

src/main/java/de/uni_tuebingen/ub/nppm/db/QuelleDB.java
```diff
@@ -15,6 +17,48 @@ public static List getList() throws Exception {
...
+                query.setParameter("zuV", 1);
+                return query.list();
+            }else{
+                Integer start = null;
+                if(currentPage != null && recordsPerPage != null)
+                    start = currentPage * recordsPerPage - recordsPerPage;
+                if(sort.startsWith("title")){
+                    q = "FROM Quelle q WHERE q.zuVeroeffentlichen = :zuV AND q.bezeichnung LIKE :bez ORDER BY q.bezeichnung "+ (sort.equals("titleDown") ? "DESC" : "ASC");
+                }else if(sort.startsWith("belege")){
+                    q = "SELECT q FROM Quelle q LEFT JOIN q.einzelbelege e WHERE q.zuVeroeffentlichen = :zuV AND q.bezeichnung LIKE :bez " +
```
mtrojan-ub: This statement seems to be very slow, especially for belegeDown. Can this be optimized?

src/main/webapp/gast/stat.jsp
```diff
@@ -0,0 +1,170 @@
...
+    public String html_page_item_current(int page) {
+        return "<li class=\"page-item active\"><a class=\"page-link\">" + page + "</a></li>";
+    }
+
+    public String html_page_item(int page, String filterTitle, int recordsPerPage, String sort, String jumpToID) {
+        return "<li class=\"page-item\"><a class=\"page-link\"href=\"?jumpToID="+jumpToID+"&sort="+sort+"&page=stat&filterTitle=" + filterTitle + "&recordsPerPage=" + recordsPerPage + "&currentPage=" + page + "\">" + page + "</a></li>";
+    }
+
+    public String html_sort_title_up(int page, String filterTitle, int recordsPerPage, String jumpToID) {
+        return "<a class=\"sort-link\"href=\"?jumpToID="+jumpToID+"&sort=titleUp&page=stat&filterTitle=" + filterTitle + "&recordsPerPage=" + recordsPerPage + "&currentPage=" + page + "\">sort up</a>";
```
mtrojan-ub: "sort up" + "sort down" should also be translated

src/main/webapp/gast/stat.jsp
```diff
@@ -0,0 +1,170 @@
...
+       <tbody>
+            <%
+            for (Quelle q : lst) {
+                out.print("<tr>");
+                out.print("<td>");
+                out.print("<a href=\""+Utils.getBaseUrl(request)+"/gast/quelle?ID="+String.valueOf(q.getId())+"\">");
+                out.print(Utils.escapeHTML(q.getBezeichnung()));
+                out.print("</a>");
+                out.print("</td>");
+                out.print("<td>");
```
mtrojan-ub: The width of the columns in the result table seems to be depending on the length of the entries in the first column depending on the selected page. Can you please give the columns a fixed width in %?

src/main/java/de/uni_tuebingen/ub/nppm/db/QuelleDB.java
juro-ub: ich habe es jetzt ohne join left und group by gemacht:

```
q = "SELECT * FROM quelle q WHERE q.zuVeroeffentlichen = :zuV AND q.bezeichnung LIKE :bez "
                            + "ORDER BY (SELECT COUNT(e.id) FROM einzelbeleg e WHERE e.QuelleID = q.id) " + (sort.equals("belegeDown") ? "DESC" : "ASC");
```

src/main/webapp/gast/stat.jsp
juro-ub: yes, fixed!

src/main/webapp/gast/stat.jsp
juro-ub: i have added a fixed percentage

src/main/java/de/uni_tuebingen/ub/nppm/db/QuelleDB.java
mtrojan-ub: Ah, ich hab herausgefunden woher die Performance-Probleme kommen, das Statement ist es nicht (siehe anderer Kommentar)

src/main/webapp/gast/stat.jsp
```diff
@@ -0,0 +1,170 @@
...
+            <%
+            for (Quelle q : lst) {
+                out.print("<tr>");
+                out.print("<td width='80%'>");
+                out.print("<a href=\""+Utils.getBaseUrl(request)+"/gast/quelle?ID="+String.valueOf(q.getId())+"\">");
+                out.print(Utils.escapeHTML(q.getBezeichnung()));
+                out.print("</a>");
+                out.print("</td>");
+                out.print("<td width='20%'>");
+                out.print(q.getEinzelbelege().size());
```
mtrojan-ub: THIS is the performance problem (not the statement). If you leave the filter empty & sort by belegeDown, then the result list will be rendered. For each Quelle there will be all einzelbelege selected & then counted (e.g. > 38.000 in Verbrüderungsbuch von Reichenau, and so on). Maybe we can provide some kind of shortcut function in the model class that provides better performance (only select the count without returning all Einzelbelege and counting them afterwards?)

juro-ub: Yes right.
I have implemented a `count function` in QuelleDB. Now the performance is much better.
(I could not implement the `count function` in the model class, because i needed access to the hibernate session)

mtrojan-ub:
Die Performance ist jetzt viel besser => ich merge es jetzt. Allerdings ist mir noch ein Bug aufgefallen => bei der Verwendung des "Filter"-Feldes bzw. dem Klicken des Buttons springt die Ansicht zur Anzeige einer einzelnen Quelle statt die Tabelle zu filtern. Bitte ggf. noch im Nachgang beheben.

PR #204, opened 2024-01-17
GND Backend with link to National Library

rnaiser:
Im Backend ist nun das Feld GND-ID , wenn eine GND - ID eingetragen wird und speichert wird ein GND Icon als Link (zur KATALOG DER DEUTSCHEN NATIONALBIBLIOTHEK)


src/main/webapp/configuration.jsp
```diff
@@ -28,6 +28,7 @@
   String txt_search = "<img src=\"layout/icons/search2.gif\" border=\"0\" alt=\"suchen\" title=\"suchen\">";  // "suchen";
   String txt_delete = "<img src=\"layout/icons/delete2.gif\" border=\"0\" alt=\"l&ouml;schen\" title=\"l&ouml;schen\">";  // "l&ouml;schen";
   String txt_newentry = "<img src=\"layout/icons/newentry2.gif\" border=\"0\" alt=\"neuer Eintrag\" title=\"neuer Eintrag\">";  // "neuer Eintrag";
+  String gnd = "<img src=\"layout/icons/gnd.png\" border=\"0\" alt=\"GND Link\" title=\"GND suchen\">";   // "GND Link"
```
mtrojan-ub: Better rename this variable to `gndIcon`. There might be cases later where `gnd` is defined as local variable, so we better try to avoid conflicts.

src/main/webapp/forms/noarray.gnd.jsp
```diff
@@ -0,0 +1,15 @@
+<%@ page import="de.uni_tuebingen.ub.nppm.db.*" isThreadSafe="false" %>
+
+<%
+    if (feldtyp.equals("gnd") && !array) {
+
+        String gndId = AbstractBase.getStringNative("SELECT " + zielAttribut + " FROM " + zielTabelle + " WHERE ID=\"" + id + "\"");
+
+        if (gndId != null && !gndId.trim().equals("")) {
+
+        String link = "<a href=\"https://portal.dnb.de/opac/simpleSearch?query=idn%3D" + gndId + "&cqlMode=true" + "\" style=\"display: inline-block; line-height: 40px; vertical-align: middle; margin-top: 10px;\"> " + gnd + "</a>";
```
mtrojan-ub: The URL should look like this:
https://explore.gnd.network/gnd/118560034

updates/33.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE person ADD COLUMN GND varchar(255) DEFAULT NULL;
+
+CREATE INDEX idx_gnd ON person (GND);
```
mtrojan-ub: We don't use the idx_ naming schema. just call the Index the same as the column (GND), that's currently the way all other indexes in the person tables are named.

updates/33.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE person ADD COLUMN GND varchar(255) DEFAULT NULL;
```
mtrojan-ub: When adding columns, please don't just add them to the end of the table. try to add them where similar fields are. In this case i suggest you add this to the statement:
`AFTER PKZ`

src/main/webapp/forms/noarray.gnd.jsp
mtrojan-ub: When generating HTML out of DB values, don't forget to escape them:
`" + gndId + "` => `" + DBtoHTML(gndId) + "`

src/main/webapp/configuration.jsp
rnaiser: done

src/main/webapp/forms/noarray.gnd.jsp
rnaiser: done

updates/33.sql
rnaiser: done

updates/33.sql
rnaiser: done

src/main/webapp/forms/noarray.gnd.jsp
rnaiser: done

src/main/webapp/person.jsp
```diff
@@ -20,445 +19,463 @@
...
-	<jsp:param name="ID" value="<%= id %>" />
-	<jsp:param name="Formular" value="person" />
-	<jsp:param name="Datenfeld" value="QuiEt" />
-</jsp:include>
-</div>
-</div>
+                            <jsp:include page="inc.erzeugeFormular.jsp">
+                                <jsp:param name="ID" value="<%=id%>" />
+                                <jsp:param name="Formular" value="person" />
+                                <jsp:param name="Datenfeld" value="GNDLink" />
```
mtrojan-ub, line 141: Do we really need to make a different call / define a different GNDLink here? I think it might be better to just generate the link when the GND field is shown.

mtrojan-ub: Ah ok now i understand ... the GND is basically a regular textfield and the GNDlink is only rendered in addition... ok then we can leave it this way, but i guess we should rename the template files to make things clearer

src/main/webapp/inc.erzeugeFormular.jsp
```diff
@@ -130,5 +130,6 @@
 <%@ include file="forms/noarray.sqlselect.jsp" %>
 <%@ include file="forms/noarray.textarea.jsp" %>
 <%@ include file="forms/noarray.textfield.jsp" %>
+<%@ include file="forms/noarray.gnd.jsp" %>
```
mtrojan-ub: Since the GND field is rendered via `noarray.textfield.jsp` and the new template is just for handling the GND links, i think we should rename it to `noarray.gndlink.jsp` (and also change the DB entry for it.)

src/main/webapp/forms/noarray.gnd.jsp
```diff
@@ -0,0 +1,15 @@
+<%@ page import="de.uni_tuebingen.ub.nppm.db.*" isThreadSafe="false" %>
+
+<%
+    if (feldtyp.equals("gnd") && !array) {
+
+        String gndId = AbstractBase.getStringNative("SELECT " + zielAttribut + " FROM " + zielTabelle + " WHERE ID=\"" + id + "\"");
+
+        if (gndId != null && !gndId.trim().equals("")) {
+
+        String link = "<a href=\"https://d-nb.info/gnd/" + DBtoHTML(gndId) + "\" style=\"display: inline-block; line-height: 40px; vertical-align: middle; margin-top: 10px;\"> " + gndIcon + "</a>";
```
mtrojan-ub: Since this is a link to an external page, we should also add `target="_blank"`.

src/main/webapp/inc.erzeugeFormular.jsp
rnaiser: ich habe es umgenannt, auch in inc.erzeugeFormular.jsp namen ersetzt, was oder wo : (and also change the DB entry for it.) ?

src/main/webapp/forms/noarray.gnd.jsp
rnaiser: done

updates/33.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE person ADD COLUMN GND varchar(255) DEFAULT NULL AFTER PKZ;
+
+CREATE INDEX GND ON person (GND);
+
+INSERT INTO datenbank_mapping (Formular, Datenfeld, de_Beschriftung, Feldtyp, Array, ZielTabelle, ZielAttribut, FormularAttribut, Seite, gb_beschriftung, fr_beschriftung, la_beschriftung) VALUES ("person", "GND", "GND-ID", "textfield", 0, "person", "GND", "PersonID", "person", "GND-ID", "GND-ID", "GND-ID");
+
+INSERT INTO datenbank_mapping (Formular, Datenfeld, Feldtyp, Array, ZielTabelle, ZielAttribut) VALUES ("person", "GNDLink", "gnd", 0, "person", "GND");
```
mtrojan-ub: Shouldn't the `Feldtyp` be `gndlink` here (same as the template name) instead of `gnd`?

rnaiser: Yes (not necessary, but yes, i did changed it and changed than the code also in noarray.gndlink  if (feldtyp.equals("gndlink") && !array) {

PR #211, opened 2024-01-31
Feld Angabe zur Person im Einzelbeleg

rnaiser:


updates/ryan_2.sql
```diff
@@ -0,0 +1,33 @@
...
+ENGINE = InnoDB
+DEFAULT CHARACTER SET = utf8mb4
+COLLATE = utf8mb4_unicode_ci;	
+
+/*Insert a default value*/
+INSERT INTO `neg`.`selektion_angabe` (ID,Bezeichnung) VALUES (-1,'-');
+INSERT INTO `neg`.`selektion_angabe` (ID,Bezeichnung) VALUES (1,'?');
+
+CREATE TABLE einzelbeleg_hatangabe (
+    ID int unsigned NOT NULL PRIMARY KEY AUTO_INCREMENT,
```
mtrojan-ub: This is basically correct, but please use the same way to describe the PRIMARY KEY as in selektion_angabe for consistency reasons.

src/main/java/de/uni_tuebingen/ub/nppm/model/SelektionAngabe.java
```diff
@@ -0,0 +1,23 @@
...
+import org.hibernate.annotations.CacheConcurrencyStrategy;
+
+@Entity
+@Table(name = "selektion_angabe")
+@Cacheable
+@org.hibernate.annotations.Cache(usage = CacheConcurrencyStrategy.READ_WRITE)
+class SelektionAngabe extends SelektionBezeichnung {
+
+    @ManyToMany(mappedBy = "angaben")
+    private Set<Einzelbeleg> einzelbeleg = new HashSet<>();
```
mtrojan-ub: Since this is a set and not a single object, we should rather call it `einzelbelege`. This also means the function should be renamed, `getEinzelbeleg` => `getEinzelbelege`. Moreover i think that we should not provide a setter function for the whole set, instead we should implement only add/remove functions as you did in the Einzelbeleg model class.

updates/ryan_2.sql
rnaiser: done

src/main/java/de/uni_tuebingen/ub/nppm/model/SelektionAngabe.java
rnaiser: done

PR #234, opened 2024-07-03
Titel Text und TitelKritik

rnaiser:


src/main/java/de/uni_tuebingen/ub/nppm/model/Einzelbeleg.java
```diff
@@ -907,4 +919,20 @@ public SelektionBeziehungGemeinschaft getBeziehungGemeinschaft() {
...
+
+    public void setTitelText(String titelText) {
+        this.titelText = titelText;
+    }
+
+    public Set<EinzelbelegHatTitelKritik> getTitelKritiken() {
+        return titelKritiken;
+    }
+
+    public void setTitelKritiken(Set<EinzelbelegHatTitelKritik> titelKritiken) {
```
mtrojan-ub: Whenever i see setter-functions like this, it always feels like we will never use it. It might make more sense to replace this function by 2 write operations like addTitelKritik() and removeTitelKritik(), which both take a single one to add/remove (also regarding the potential logic in a JSP template).

What do you think?

src/main/java/de/uni_tuebingen/ub/nppm/model/EinzelbelegHatTitelKritik.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @ManyToOne(targetEntity = SelektionTitelKritik.class)
+    @JoinColumn(name = "TitelkritikID", referencedColumnName = "ID")
+    private SelektionTitelKritik titelKritik;
+
+    public Integer getId() {
+        return id;
+    }
+
+    public void setId(Integer id) {
```
mtrojan-ub: We will never need a setId() function because this will always be auto-generated via the database via AUTO_INCREMENT.

src/main/java/de/uni_tuebingen/ub/nppm/model/SelektionTitelKritik.java
```diff
@@ -0,0 +1,31 @@
...
+@Entity
+@Table(name = "selektion_titelkritik")
+@Cacheable
+@org.hibernate.annotations.Cache(usage = CacheConcurrencyStrategy.READ_WRITE)
+public class SelektionTitelKritik extends SelektionBezeichnung{
+
+     @ManyToMany(mappedBy = "titelKritiken")
+    private Set<Einzelbeleg> einzelbeleg = new HashSet<>();
+
+    public Set<Einzelbeleg> getEinzelbeleg() {
```
mtrojan-ub: Since we get a List / Set, we should name the function getEinzelbelege(). The variable should be named as plural form as well.

src/main/java/de/uni_tuebingen/ub/nppm/model/SelektionTitelKritik.java
```diff
@@ -0,0 +1,31 @@
...
+        return einzelbeleg;
+    }
+
+    public void setEinzelbeleg(Set<Einzelbeleg> einzelbeleg) {
```
mtrojan-ub: Here as well, i dont think we ever need this setter and should switch to add/remove single ones to the list.

src/main/java/de/uni_tuebingen/ub/nppm/model/SelektionTitelKritik.java
```diff
@@ -0,0 +1,31 @@
...
+        this.einzelbeleg = einzelbeleg;
+    }
+
+     public void addEinzelbeleg(Einzelbeleg beleg) {
```
mtrojan-ub, line 20: Please watch your indentations, seems to be wrong in multiple occasions throughout the code.

src/main/java/de/uni_tuebingen/ub/nppm/model/EinzelbelegHatTitelKritik.java
rnaiser: kann man drin lassen, falls in zukunft benötigt, eigentlich nicht, aber schaden tut es auch nicht ... ist es ok wenn ichs drin lasse ?

src/main/java/de/uni_tuebingen/ub/nppm/model/SelektionTitelKritik.java
rnaiser: done

src/main/java/de/uni_tuebingen/ub/nppm/model/SelektionTitelKritik.java
rnaiser: done

src/main/java/de/uni_tuebingen/ub/nppm/model/Einzelbeleg.java
rnaiser: done

src/main/java/de/uni_tuebingen/ub/nppm/model/SelektionTitelKritik.java
rnaiser: done

src/main/java/de/uni_tuebingen/ub/nppm/model/EinzelbelegHatTitelKritik.java
mtrojan-ub: Bitte mach's lieber raus, denn wenn man es drin lässt und es "versehentlich" mal jemand benutzt wird das bei den ganzen AUTO_INCREMENT spalten die wir haben ziemlich sicher Fehler produzieren.

rnaiser: ok entfernt .. done

PR #239, opened 2024-08-02
Fixes selektion arealtyp

juro-ub:
-    einzelbeleg_hatareal hat zwar ein Feld aber keinen Fremdschlüssel auf selektion_arealtyp
-    Außerdem sollten bei allen neuen Selektionen auch gleich die Provenance Felder mit angelegt werden
 -   Wenn ein selektion_arealtyp eingeführt wird sollte dazu auch eine passende Model Class angelegt werden


src/main/java/de/uni_tuebingen/ub/nppm/model/SelektionArealTyp.java
```diff
@@ -0,0 +1,27 @@
...
+
+    public void addEinzelbeleg(Einzelbeleg beleg) {
+        this.getEinzelbeleg().add(beleg);
+    }
+
+    public void removeEinzelbeleg(int id) {
+        this.getEinzelbeleg().removeIf(e -> e.getId() == id);
+    }
+
+    public Set<Einzelbeleg> getEinzelbeleg() {
```
mtrojan-ub: Basically OK, but as also discussed with @rnaiser regarding other PRs:
- If a function returns a Set, it should be called by the plural form => getEinzelbelege
- It almost never makes sense to have a setEinzelbelege function called with a set. Instead it's always better to have single addEinzelbeleg and removeEinzelbeleg functions.

updates/julian_2.sql
```diff
@@ -9,11 +9,16 @@ VALUES ('einzelbeleg','NewAreal','combined',1,'einzelbeleg_hatareal','ArealID;Ar
 
 ALTER TABLE `einzelbeleg_hatareal` ADD COLUMN `ArealTypID` int DEFAULT -1;
 
+ALTER TABLE einzelbeleg_hatareal ADD CONSTRAINT fk_selektion_arealtyp FOREIGN KEY (ArealTypID) REFERENCES selektion_arealtyp(ID);
```
mtrojan-ub: Please stick to the proper naming schema for our foreign keys:
- einzelbeleg_hatareal_ArealTypID

updates/julian_2.sql
```diff
@@ -9,11 +9,16 @@ VALUES ('einzelbeleg','NewAreal','combined',1,'einzelbeleg_hatareal','ArealID;Ar
...
+
 CREATE TABLE `selektion_arealtyp` (
   `ID` int NOT NULL AUTO_INCREMENT,
   `Bezeichnung` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci DEFAULT NULL,
+  `provenance_source` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT('NeG'),
+  `provenance_id` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci DEFAULT NULL,
   PRIMARY KEY (`ID`),
-  UNIQUE KEY `Bezeichnung` (`Bezeichnung`)
+  UNIQUE KEY `uk_bezeichnung` (`Bezeichnung`),
+  UNIQUE KEY `uk_provenance_id` (`provenance_id`)
```
mtrojan-ub: Please also stick to a naming scheme regarding UNIQUE KEYs. In the other PR you called them unique_provenance_id. Also please find a proper unique key name for Bezeichnung.

src/main/java/de/uni_tuebingen/ub/nppm/model/SelektionArealTyp.java
juro-ub: fixed!

updates/julian_2.sql
juro-ub: fixed!

updates/julian_2.sql
juro-ub: fixed!

src/main/java/de/uni_tuebingen/ub/nppm/model/SelektionArealTyp.java
```diff
@@ -7,21 +7,17 @@
...
+    private Set<Einzelbeleg> einzelbelege = new HashSet<>();
 
     public void addEinzelbeleg(Einzelbeleg beleg) {
-        this.getEinzelbeleg().add(beleg);
+        this.getEinzelbelege().add(beleg);
     }
 
     public void removeEinzelbeleg(int id) {
-        this.getEinzelbeleg().removeIf(e -> e.getId() == id);
+        this.getEinzelbelege().removeIf(e -> e.getId() == id);
```
mtrojan-ub: Einzelbeleg.getId() returns an Integer, so comparing with == will always be false. You should use .equals() instead

juro-ub: i have changed the comparison to equals()

updates/julian_2.sql
```diff
@@ -9,11 +9,16 @@ VALUES ('einzelbeleg','NewAreal','combined',1,'einzelbeleg_hatareal','ArealID;Ar
 
 ALTER TABLE `einzelbeleg_hatareal` ADD COLUMN `ArealTypID` int DEFAULT -1;
 
+ALTER TABLE einzelbeleg_hatareal ADD CONSTRAINT einzelbeleg_hatareal_ArealTypID FOREIGN KEY (ArealTypID) REFERENCES selektion_arealtyp(ID);
```
mtrojan-ub: This foreign key cannot be added before the new table selektion_arealtyp is created, so it should be moved behind the corresponding CREATE TABLE statement.

Also, please keep a certain normalized syntax at least within the same SQL file => in all other statements, table + column names are wrapped with backticks.

juro-ub: fixed!

PR #240, opened 2024-08-06
Rename MGHLemma, Philologisches Lemma

juro-ub:
Rename:

MGHLemma => Lemma
Philologisches Lemma => Philologischer Kommentar
Indexlemma => bleibt so


mtrojan-ub:
As discussed => since the names are not really final yet, we should put this topic "on hold" until the Lemma workshop at the end of this year.

juro-ub:
> As discussed => since the names are not really final yet, we should put this topic "on hold" until the Lemma workshop at the end of this year.

I have now implemented the changes from the workshop

PR #244, opened 2024-08-26
DB cleanup scripts

mtrojan-ub:
@rnaiser: You asked for examples regarding cleanup scripts / CSV => SQL converters, so here they are.


mtrojan-ub:
Note: funktion.php is still work in progress, however it looks like the script works but executing the created SQL file is not problems due to a problem in Selektion_Funktion.csv => Column D seems to be wrong + C seems to be empty for IDs (Column B) 504+507. I already asked our stakeholder to correct the data.

PR #251, opened 2024-09-26
inc.modul.jsp erneuert mit hibernate

rnaiser:


src/main/webapp/inc.modul.jsp
```diff
@@ -1,1772 +1,1470 @@
...
+                }
+                out.print(bezeichnung);
+                first = false;
+            } //while 2 ende
+            out.println("</td>");
+            out.println("</tr>");
+        } //while 1 ende
+
+        if (firstEdition) {
+            out.println("<td></td><td></td><td></td><td></td><td></td><td></td><td></td><td></td></tr>");
```
mtrojan-ub: I know it has been like this before, but maybe this would be a good opportunity to optimize:
`<td></td><td></td><td></td><td></td><td></td><td></td><td></td><td></td>`
is equivalent to
`<td colspan="8"></td>`

src/main/webapp/inc.modul.jsp
```diff
@@ -1,1772 +1,1470 @@
...
+                int count = 0;
+                for (Object[] row : resultList) {
+                    count++;
+                    if (count % 2 == 0) {
+                        out.println("<tr>");
+                    } else {
+                        out.println("<tr bgcolor='#AACCDD'>");
+                    }
+
+                    String zitierweise = row[0] != null ? String.valueOf(row[0]) : "";
```
mtrojan-ub, line 56: Before this PR, it was possible to access columns in the ResultSet via the proper column name (e.g. "ed.Zitierweise"), but now they are accessed via a counter. The queries used in the ModulIncDB are very complex with tons of different columns. If the query is ever changed, the counter will also change which is very dangerous. Can we find any solution to keep the addressing via column name here?

src/main/webapp/inc.modul.jsp
rnaiser: done

src/main/webapp/inc.modul.jsp
rnaiser: Nein, der Counter hat damit nichts zu tun; er ist nur dafür da, unterschiedliche Hintergrundfarben anzuzeigen, und im neuen Design wird der Counter sowieso entfernt.

mtrojan-ub: As discussed => no, this is related to the columns inside the sql result row. It seems to be a lot of effort to correct this now, but since it is only used as single select query in the module it might be locally enough to keep it like this. However, i'd strongly prefer to keep the by-string accessor for everything else which is coded in the future.

mtrojan-ub:
Lots of changes in here, but since it is necessary to change it due to the hibernate connection provider & we're about to start a major test phase on our test server, we better merge it now than later. I just hope you tested everything really extensively.

PR #253, opened 2024-10-01
Add Provenance information + Extended Search

juro-ub:
- show provenance information for Einzelbeleg, Quellen, Namen, Lemma
- Show provenance information for vocabublaries
- Extend backend search to filter by provenance src

### DIFF
--- a/src/main/java/de/uni_tuebingen/ub/nppm/db/AbstractBase.java
+++ b/src/main/java/de/uni_tuebingen/ub/nppm/db/AbstractBase.java
@@ -498,7 +498,7 @@ public class AbstractBase {
         return getSingleField("provenance_id", tabelle, Integer.valueOf(id));
     }
 
-    public static String getProvenanceSrc(String id, String tabelle) throws Exception{
+    public static String getProvenanceSource(String id, String tabelle) throws Exception{
         return getSingleField("provenance_source", tabelle, Integer.valueOf(id));
     }
 }
--- a/src/main/java/de/uni_tuebingen/ub/nppm/db/AbstractBase.java
+++ b/src/main/java/de/uni_tuebingen/ub/nppm/db/AbstractBase.java
@@ -493,4 +493,12 @@ public class AbstractBase {
             throw new SqlInjectionException("Invalid column name: " + column);
         }
     }
+
+    public static String getProvenanceId(String id, String tabelle) throws Exception{
+        return getSingleField("provenance_id", tabelle, Integer.valueOf(id));
+    }
+
+    public static String getProvenanceSrc(String id, String tabelle) throws Exception{
+        return getSingleField("provenance_source", tabelle, Integer.valueOf(id));
+    }
 }
--- a/src/main/java/de/uni_tuebingen/ub/nppm/db/SelektionDB.java
+++ b/src/main/java/de/uni_tuebingen/ub/nppm/db/SelektionDB.java
@@ -29,6 +29,14 @@ public class SelektionDB extends AbstractBase {
         return getList(getEntityClassByTableName(selektion));
     }
 
+    static public List<SelektionBezeichnung> getListBezeichnung(String selektion) throws Exception {
+        return getList(getEntityClassByTableName(selektion));
+    }
+
+    static public List<SelektionProvenance> getListProvenance(String selektion) throws Exception {
+        return getList(getEntityClassByTableName(selektion));
+    }
+
     static public List<SelektionHierarchy> getListHierarchy(String selektion) throws Exception {
         Class c = getEntityClassByTableName(selektion);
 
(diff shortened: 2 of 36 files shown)


src/main/java/de/uni_tuebingen/ub/nppm/db/AbstractBase.java
```diff
@@ -493,4 +493,12 @@ public static void verifyDynamicColumn(String column) throws SqlInjectionExcepti
...
         }
     }
-}
\ No newline at end of file
+
+    public static String getProvenanceId(String id, String tabelle) throws Exception{
+        return getSingleField("provenance_id", tabelle, Integer.valueOf(id));
+    }
+
+    public static String getProvenanceSrc(String id, String tabelle) throws Exception{
```
mtrojan-ub: We don't really use abbreviations in the base yet, so i'd prefer renaming the function to getProvenanceSource.

juro-ub: fixed!

src/main/java/de/uni_tuebingen/ub/nppm/model/SelektionProvenance.java
```diff
@@ -0,0 +1,31 @@
+package de.uni_tuebingen.ub.nppm.model;
+
+import javax.persistence.*;
+import org.hibernate.annotations.CacheConcurrencyStrategy;
+
+@MappedSuperclass
+@Cacheable
+@org.hibernate.annotations.Cache(usage = CacheConcurrencyStrategy.READ_WRITE)
+public abstract class SelektionProvenance extends SelektionBezeichnung {
```
mtrojan-ub, line 9: It's basically OK to introduce a new class, but since we really want to have provenance on all types of selections (e.g. to show them all in the Administration backend) we might also just add the new fields directly to Selektion.java.

src/main/webapp/forms/id.jsp
```diff
@@ -1,28 +1,49 @@
...
     }
 
     out.println("ID: ");
     if (title.toLowerCase().equals("einzelbeleg")) {
         out.print("B");
+        provenanceId = EinzelbelegDB.getProvenanceId(id, "einzelbeleg");
+        provenanceSrc = EinzelbelegDB.getProvenanceSource(id, "einzelbeleg");
     } else if (title.toLowerCase().equals("person")) {
         out.print("P");
+        provenanceId = EinzelbelegDB.getProvenanceId(id, "person");
```
mtrojan-ub: The getProvenanceId method is implemented in AbstractBase. You should not try to use the EinzelbelegDB to call methods on entities for which it's not responsible. So please either use PersonDB, NamenKommentarDB... and so on here, or call AbstractBase directly. There are some other cases i already discussed with @rnaiser where we call AbstractBase (even though we shouldn't since the main idea was just having it as template for the subclasses) => There is a certain possibility that we might either rename AbstractBase to Base, or maybe add an intermediate GenericDB layer for these kinds of methods in the future.

src/main/java/de/uni_tuebingen/ub/nppm/db/SelektionDB.java
```diff
@@ -72,6 +80,54 @@ static public SelektionBezeichnung getByBezeichnung(String selektion, String Bez
...
+            String sql = "SELECT Bezeichnung FROM " + tabelle + " WHERE ID= :ID";
+            NativeQuery query = session.createNativeQuery(sql);
+            query.setParameter("ID", id);
+            result = (String) query.getSingleResult();
+            session.getTransaction().commit();
+        }
+        return result;
+    }
+
+    public static String getProvenanceSourceByID(String id, String tabelle) throws Exception {
```
mtrojan-ub: Isn't this basically exactly the same as AbstractBase.getProvenanceSource()?

juro-ub: Yes i have removed the duplicate

src/main/webapp/forms/id.jsp
juro-ub: Yes, i have used now the specific db classes like PersonDB, NamenkommentarDB, MGHLemmaDB...

src/main/java/de/uni_tuebingen/ub/nppm/model/SelektionProvenance.java
juro-ub: The problem is here that there are some Selektions Entities with no provenance information. Like SelektionAmtWeiheGastEinzelbeleg , SelektionAmtWeiheGastPerson and SelektionQuellengattungGast...

mtrojan-ub: Ah, alright! Then let's just keep it this way, thanks!

PR #284, opened 2025-03-12
Rest API Lemma Support

juro-ub:
Neue Aufruf Möglichkeit:

http://localhost:8080/neg/rest/lemmas/belegform1,belegform2
http://localhost:8080/neg/rest/lemma/belegform1

Rückgabe:

Lemma ID+Lemma

### DIFF
--- a/src/main/java/de/uni_tuebingen/ub/nppm/db/LemmaDB.java
+++ b/src/main/java/de/uni_tuebingen/ub/nppm/db/LemmaDB.java
@@ -54,18 +54,18 @@ public class LemmaDB extends AbstractBase {
         return getStringListNative("SELECT DISTINCT SUBSTRING_INDEX(MGHLemma, '~', -1) AS Zweitglied  FROM neg.mgh_lemma WHERE MGHLemma LIKE '%~%' ORDER BY Zweitglied ASC");
     }
 
-    public static MghLemma getLemmaByBelegform(String belegform) throws Exception {
+    public static List<MghLemma> getLemmaByBelegform(String belegform) throws Exception {
         String sql = "SELECT ml.* FROM mgh_lemma ml "
                 + "JOIN einzelbeleg_hatmghlemma ehm ON ml.ID = ehm.MGHLemmaID "
                 + "JOIN einzelbeleg eb ON ehm.EinzelbelegID = eb.ID "
                 + "WHERE eb.Belegform = :belegform "
-                + "ORDER BY ml.MGHLemma "
-                + "LIMIT 1";
+                + "ORDER BY ml.MGHLemma ";
 
         try (Session session = getSession()) {
             NativeQuery<MghLemma> sqlQuery = session.createNativeQuery(sql, MghLemma.class);
             sqlQuery.setParameter("belegform", belegform);
-            return sqlQuery.uniqueResultOptional().orElse(null);
+            List<MghLemma> results = sqlQuery.getResultList();
+            return results; // Genau ein Treffer
         }
     }
 }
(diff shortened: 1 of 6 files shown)


src/main/java/de/uni_tuebingen/ub/nppm/db/LemmaDB.java
```diff
@@ -53,4 +53,19 @@ public static List<String> getListErstglied() throws Exception {
...
         return getStringListNative("SELECT DISTINCT SUBSTRING_INDEX(MGHLemma, '~', -1) AS Zweitglied  FROM neg.mgh_lemma WHERE MGHLemma LIKE '%~%' ORDER BY Zweitglied ASC");
     }
+
+    public static MghLemma getLemmaByBelegform(String belegform) throws Exception {
+        String sql = "SELECT ml.* FROM mgh_lemma ml "
+                + "JOIN einzelbeleg_hatmghlemma ehm ON ml.ID = ehm.MGHLemmaID "
+                + "JOIN einzelbeleg eb ON ehm.EinzelbelegID = eb.ID "
+                + "WHERE eb.Belegform = :belegform "
+                + "ORDER BY ml.MGHLemma "
+                + "LIMIT 1";
```
mtrojan-ub: Careful with LIMIT 1 => if there is more than 1 lemma for this belegform we explicitly want to return some kind of error

src/main/java/de/uni_tuebingen/ub/nppm/servlet/RESTServlet.java
```diff
@@ -89,6 +95,16 @@ private void processMultipleIdentifiers(String[] identifiers, HttpServletRespons
                     if (einzelbeleg != null) {
                         jsonObject = einzelbeleg.getJSON();
                     }
+                } else if(id.startsWith("P")) {
+                    Person person = (Person) IdentifierMapper.getModelByIdentifier(id);
```
mtrojan-ub, line 106: Will there be any error reporting if there is no person or quelle with this ID?

src/main/webapp/resthelp.jsp
```diff
@@ -74,57 +86,71 @@
...
+    <p>URL: <code>/rest/lemma/Sebastianus</code></p>
+    <pre>
+Response:
+{
+  "ID": "M6360",
+  "Lemma": "Sebastianus"
+}
+    </pre>
+
+    <h3>Fetch Multiple Lemmas by Belegformen</h3>
```
mtrojan-ub: On these kind of documentation pages we should really make sure that, if they are in english, the proper english terms are used.

"Lemma" is OK, but for "Belegform" we should use something like "Reference Form" (e.g. if you switch the frontend to english, "Einzelbeleg" is "Single Reference" and "Belegform" is like "Form of Reference", and so on.

mtrojan-ub:
.. and we should add a hint in resthelp.jsp (below "welcome to the REST API for the NPPM project"):

`Please note that this API is still experimental, which means that the listed endpoints and the returned data structures can potentially change at any time. If there are specific uses cases that you would like to cover in a stable way, please contact us at nppm-team@ub.uni-tuebingen.de.`

(this should be in some kind of box with a warning symbol)

src/main/java/de/uni_tuebingen/ub/nppm/db/LemmaDB.java
juro-ub: i have removed the LIMIT 1 from the sql and added a error handling in the RESTServlet if there are no elements or more than 1 element

src/main/java/de/uni_tuebingen/ub/nppm/servlet/RESTServlet.java
juro-ub: I have Added a Error Handling for the different items

juro-ub:
> .. and we should add a hint in resthelp.jsp (below "welcome to the REST API for the NPPM project"):
> 
> `Please note that this API is still experimental, which means that the listed endpoints and the returned data structures can potentially change at any time. If there are specific uses cases that you would like to cover in a stable way, please contact us at nppm-team@ub.uni-tuebingen.de.`
> 
> (this should be in some kind of box with a warning symbol)

I have added the warning box

PR #289, opened 2025-03-21
REST API fix

juro-ub:
- Use UTF-8 in Helppage
- New Examples
- dont use array for single item queries
- Refactor Error Handling
- Include Error Handling in resthelp page

### DIFF
--- a/src/main/webapp/resthelp.jsp
+++ b/src/main/webapp/resthelp.jsp
@@ -289,8 +289,9 @@ Response:
                             "mghLemma": "leub-n",
                             "id": "M6361"
                         },
-                        {
-                            "error": "Lemma not found with ID M9999999"
+                        {                            
+                            "error": "Lemma not found with ID M9999999",
+                            "id": "M9999999"
                         }
                     ]
                 }
--- a/src/main/webapp/resthelp.jsp
+++ b/src/main/webapp/resthelp.jsp
@@ -240,7 +240,7 @@ Response:
         <td>500 (Internal Server Error)</td>
         <td>Internal Server Error: {message}</td>
         <td>An unexpected error occurred while processing the request.</td>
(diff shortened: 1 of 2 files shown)


src/main/webapp/resthelp.jsp
```diff
@@ -118,64 +118,162 @@
...
+        <td>500 (Internal Server Error)</td>
+        <td>Internal server error due to malformed item identifiers</td>
+        <td>The URL path does not match the expected format for multiple items (e.g., missing commas or invalid characters).</td>
+        <td><code>/rest/items/M6361M6362</code></td>
+    </tr>
+    <tr>
+        <td>500 (Internal Server Error)</td>
+        <td>Internal Server Error: {message}</td>
+        <td>An unexpected error occurred while processing the request.</td>
+        <td><code>/rest/items/M6361,M6362</code> (if database is down)</td>
```
mtrojan-ub, line 275: There should also be an example what happens when you e.g. use the items endpoint to request 2 items and 1 of them cannot be found (including example JSON result).

juro-ub: I have added an Error Example with JSON Response

src/main/webapp/resthelp.jsp
```diff
@@ -118,64 +118,188 @@
...
+        <td>
+            <pre>
+                {
+                    "items": [
+                        {
+                            "mghLemma": "leub-n",
+                            "id": "M6361"
+                        },
+                        {
+                            "error": "Lemma not found with ID M9999999"
```
mtrojan-ub: In case of error the "id" field should be there as well for better error handling on the client side.

juro-ub: i have added the id field

PR #334, opened 2025-05-06
Fix Navigation Start Button

### LINKED ISSUE #332: Startseite nicht in Navigation hervorgehoben wenn aktiv
Siehe https://github.com/ubtue/neg/pull/330#issuecomment-2850510808

Auf der Startseite sollte in der Navigation der Tab "Start" hervorgehoben sein wenn die Seite aktiv ist, so wie auf allen anderen Unterseiten auch der entsprechende Tab hervorgehoben ist.

Das ist zwar ein Usability Thema aber laut @rnaiser schwierig umzusetzen, daher die Bitte an @juro-ub ob Du Dir das anschauen kannst.

juro-ub:
Highlight start in navigation.
Solved: override getDynamicNavigationTitle(...)

### DIFF
--- a/src/main/java/de/uni_tuebingen/ub/nppm/servlet/LoginServlet.java
+++ b/src/main/java/de/uni_tuebingen/ub/nppm/servlet/LoginServlet.java
@@ -66,7 +66,7 @@ public class LoginServlet extends HttpServlet {
 
         // Weiterleiten
         if (benutzer.isGast()) {
-            response.sendRedirect(Utils.getBaseUrl(request) + "/gast/infos?sharedHtml=start");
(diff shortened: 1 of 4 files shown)


mtrojan-ub:
Basically works, but: If you access the page from the entry point (e.g. localhost/neg/) then the redirect will have no "current" parameter so that should also be added (however i still think the solution regarding the "current" in the url is ugly and should be replaced in the long run as described in #297)

juro-ub:
> Basically works, but: If you access the page from the entry point (e.g. localhost/neg/) then the redirect will have no "current" parameter so that should also be added (however i still think the solution regarding the "current" in the url is ugly and should be replaced in the long run as described in #297)

I have fixed this case

mtrojan-ub:
related to issue #332

PR #335, opened 2025-05-06
Issue 318: Lemmas mit [???] im Backend behalten aber im Frontend verstecken

### LINKED ISSUE #318: Lemmas mit [???] im Backend behalten aber im Frontend verstecken
Beispiel:
`athal~[???]`

IDs bezieht sich auf aktuellen Stand des Testservers, falls erneut migriert wird ändern sie sich gegebenenfalls.

Nach Änderung:
- Öffnen per ID-Feld zu M7452 springt jetzt zum nächsten verfügbaren Lemma M7453
- Öffnen per URL wirft jetzt den Fehler `Lemma ID M7452 enthält ungültige Zeichenkette -> '[???]' `

juro-ub:
[Lemmas mit [???] im Backend behalten aber im Frontend verstecken]
(https://github.com/ubtue/neg/issues/318)


src/main/webapp/gast/lemma.jsp
```diff
@@ -30,22 +30,34 @@
...
             } else{
                 throw new IdNotFoundException("Lemma ID M" + String.valueOf(id) + " is not to be published");
             }
         }
+
+        if (throwContainsInvalidStrException) {
+            if (session.getAttribute("Sprache").equals("de")) {
+                throw new ContainsInvalidStrException("Lemma ID M" + id + " enthält ungültige Zeichenkette -> '[???]'");
+            } else{
+                throw new ContainsInvalidStrException("Lemma ID M" + String.valueOf(id) + " contains invalid string -> '[???]'");
```
mtrojan-ub: It's OK like this right now, but for the long perspective we should also store these kind of error messages in the database.

mtrojan-ub:
I can still open Lemma 8653 in the frontend even if it contains [???]

mtrojan-ub:
Lemmas with [???] are also still listed in the autocomplete function as well as the search result when using the advanced search

juro-ub:
> I can still open Lemma 8653 in the frontend even if it contains [???]

I am not able to reproduce this problem. If i try to open the Lemma with the ID 8653 via the ID-Field, i get the following error:
image

Can you please tell me how you tried to open the Lemma with ID 8653

PS:
the error will only be shown if you try to open the lemma directly via the url or the "ID Field".
In the pagianation the lemmas with [???] will not be considered

juro-ub:
> ed in the autocomplete function

I have implemented the functionality. Now the autocomplete and the extended search will hide MGHLemma-entries with "[???]"

But i found out that there are also entries like

- "[???"
- "[???)"
- "???"

These entries will not be hidden only the entries that include "[???]"

mtrojan-ub:
> Can you please tell me how you tried to open the Lemma with ID 8653

Via URL

Note: There is a certain logic that jumps to the next public ID of an entity type if the requested entity is not public. I think it was implemented 1-2 weeks ago by @rnaiser when you use the ID form in the navigation. Basically we should use the same logic here, and also make sure that there is a HTTP redirect that includes response code 303 so that crawlers/bots know that the response is actually not the entity that they requested, instead of showing the "Lemma contains incorrect string" error.

juro-ub:
> > Can you please tell me how you tried to open the Lemma with ID 8653
> 
> Via URL
> 
> Note: There is a certain logic that jumps to the next public ID of an entity type if the requested entity is not public. I think it was implemented 1-2 weeks ago by @rnaiser when you use the ID form in the navigation. Basically we should use the same logic here, and also make sure that there is a HTTP redirect that includes response code 303 so that crawlers/bots know that the response is actually not the entity that they requested, instead of showing the "Lemma contains incorrect string" error.

Yes i will implement the response code 303.
I asked ryan about the logic.
If the id is passed via url there should be a error message for the user and if the id is passed via the id input field, the logic should jump to the next valid mgh_lemma.
I will it implement like this

Edit:
Its implemented now

PR #344, opened 2025-05-14
pagination (titel.inc.jsp) + Ergebnisfelder

rnaiser:
die Pagination für Gast Einzelbeleg, Personen, Namen und Quelle  also die title.inc  wurde modernisiert.

Betroffene Issues:
#307
#301

PR #345, opened 2025-05-14
First+Last Button in Statistic and Translation

juro-ub:
- Add First and Last Button
- Add Translation for Buttons
- Use of baseURL Function
- Outsource some pagination functions to reuse it for other statistic sites in the future
- Use same translation texts for search and statistic pagination

### DIFF
--- a/updates/julian_1.sql
+++ b/updates/julian_1.sql
@@ -1,0 +1,4 @@
+INSERT INTO `neg`.`datenbank_texte` (`Formular`, `Textfeld`, `de`, `gb`, `fr`, `la`) VALUES ('stat', 'Next', 'Weiter', 'Next', 'Suivant', 'Deinde');
+INSERT INTO `neg`.`datenbank_texte` (`Formular`, `Textfeld`, `de`, `gb`, `fr`, `la`) VALUES ('stat', 'Prev', 'Zurück', 'Previous', 'Précédente', 'Praecedens');
+INSERT INTO `neg`.`datenbank_texte` (`Formular`, `Textfeld`, `de`, `gb`, `fr`, `la`) VALUES ('stat', 'First', 'Anfang', 'First', 'D\'abord', 'Primus');
+INSERT INTO `neg`.`datenbank_texte` (`Formular`, `Textfeld`, `de`, `gb`, `fr`, `la`) VALUES ('stat', 'Last', 'Ende', 'Last', 'Dernière', 'Ultimus');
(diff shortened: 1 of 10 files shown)


mtrojan-ub:
Are there any potential conflicts between this PR & #344?

juro-ub:
> Are there any potential conflicts between this PR & #344?

Both PR are modifying QuelleDB.java.
PR 344 is adding the function getAllPublicQuellenIds(...)
PR 345 is adding the function getList(...)

This should be solveable.

Further we use different paginations here. Stat is using an List-view to display multiple entries on one page.
And the existing pagination shows only one item in a detail view, offering next and previous function.

The list overview in the stat page is a nice thing, because the user can sort entries and has better option to navigate through it.

In the future we can put these two pagination concepts together, but i think for now it is better to use both concepts,
because the merging of these will be complex

updates/julian_1.sql
```diff
@@ -0,0 +1,4 @@
+INSERT INTO `neg`.`datenbank_texte` (`Formular`, `Textfeld`, `de`, `gb`, `fr`, `la`) VALUES ('stat', 'Next', 'Weiter', 'Next', 'Suivant', 'Deinde');
```
mtrojan-ub: Pagination will be used on several sub pages and should always have the same texts (even if the mechanism might be different). So i suggest we rather store the texts e.g. as Formular = 'pagination' instead of 'stat' and re-use them later in all other cases.

mtrojan-ub:
You're right about the different mechanisms... i guess that browsing single pages is something different because it's more like a navigation (or browsing) than a pagination. Maybe we can use the same mechanism e.g. in the search result? On the other hand, the search result is so specific because of the displayed hierarchies, groups etc. that we might even have at least 3 different scenarios. So i think we can definitely keep the separate solutions for now, me might come back to this topic in the next funding phase when we introduce the Solr search which also needs some kind of pagination.

However, as mentioned in the other comment, it would be good to at least share the display texts if possible.

updates/julian_1.sql
mtrojan-ub: (that is, only if the mechanism would actually allow it, i just noticed there are also duplicates in the existing pages for texts like Lemma and so on.)

juro-ub: > (that is, only if the mechanism would actually allow it, i just noticed there are also duplicates in the existing pages for texts like Lemma and so on.)

I am not sure the mechanism allows it. For now we specify the translation for each Textfield in a Formular.
If we remove duplicates and use a Textfield in many Formulars, it will be not easy to maintain the translation in the future, because you dont know which formular is affected when you want to change a Textfield.

But i think for the pagination it is fine, because its really a general component in many Formulars.
I will do it for the pagination in this PR
Maybe we can identify general components like the pagination and use a general translation that is used by all formulars, in an other PR

juro-ub:
> You're right about the different mechanisms... i guess that browsing single pages is something different because it's more like a navigation (or browsing) than a pagination. Maybe we can use the same mechanism e.g. in the search result? On the other hand, the search result is so specific because of the displayed hierarchies, groups etc. that we might even have at least 3 different scenarios. So i think we can definitely keep the separate solutions for now, me might come back to this topic in the next funding phase when we introduce the Solr search which also needs some kind of pagination.
> 
> However, as mentioned in the other comment, it would be good to at least share the display texts if possible.

Yes, the pagination in the "gast search" and in the "gast statistic" are using the same translation now

juro-ub:
> > Are there any potential conflicts between this PR & #344?
> 
> Both PR are modifying QuelleDB.java. PR 344 is adding the function getAllPublicQuellenIds(...) PR 345 is adding the function getList(...)
> 
> This should be solveable.
> 
> Further we use different paginations here. Stat is using an List-view to display multiple entries on one page. And the existing pagination shows only one item in a detail view, offering next and previous function.
> 
> The list overview in the stat page is a nice thing, because the user can sort entries and has better option to navigate through it.
> 
> In the future we can put these two pagination concepts together, but i think for now it is better to use both concepts, because the merging of these will be complex

Edit: Branch has conflicts now with #344

PR #348, opened 2025-05-14
Einfache Suche Query optimieren

juro-ub:
Kleine Änderung am Select


mtrojan-ub:
Thanks! That makes the query more straightforward, but it doesn't seem like it's getting faster. I will merge it anyway.

PR #349, opened 2025-05-15
Some translations

juro-ub:


src/main/webapp/gast/einzelbeleg.jsp
```diff
@@ -53,19 +53,15 @@
     Einzelbeleg einzelbeleg = EinzelbelegDB.getById(id);
 
         if(einzelbeleg == null){
-            if (session.getAttribute("Sprache").equals("de")) {
-                throw new IdNotFoundException("Einzelbeleg ID B" + String.valueOf(id) + " ist nicht vorhanden");
-            } else{
-                throw new IdNotFoundException("Single Reference ID B" + String.valueOf(id) + " does not exist");
-            }
+            String msg = DatenbankDB.getLabel(session.getAttribute("Sprache").toString(),"einzelbeleg", "IdNotFoundError");
+            msg = msg.replace("###ID###", String.valueOf(id));
```
mtrojan-ub: There are existing texts in the database that use `%s` as placeholder:
`SELECT id, de FROM datenbank_texte WHERE de LIKE '%\%s%';`

Please use the same mechanism. It might be a good thing to add another optional parameter to the getLabel() that allows to pass a list of strings that should be replaced, and that method to execute the replacement internally.

mtrojan-ub: (However i have to add that i dont like the existing mechanism entirely, because sometimes you need to switch the sequence of placeholders in other languages, and when %s is used for multiple placeholders then you can just enter a list, so named parameters would be better - but i think it should be ok to leave the %s mechanism as-is for now and change it whenever needed in the future.)

juro-ub: yes, i changed the placeholder variable and created a getLabel function that is assigning the value

updates/julian_3.sql
```diff
@@ -0,0 +1,9 @@
+INSERT INTO `neg`.`datenbank_texte` (`Formular`, `Textfeld`, `de`, `gb`, `fr`, `la`) VALUES ('einzelbeleg', 'IdNotFoundError', 'Einzelbeleg ID B%s ist nicht vorhanden', 'Single Reference ID B%s does not exist', 'L\'ID de référence unique B%s n\'existe pas', 'Identificatio referentialis singularis B%s non exstat');
+INSERT INTO `neg`.`datenbank_texte` (`Formular`, `Textfeld`, `de`, `gb`, `fr`, `la`) VALUES ('einzelbeleg', 'NotPublicError', 'Einzelbeleg ID B%s ist nicht zu veröffentlichen', 'Single Reference ID B%s is not to be published', 'L\'ID de référence unique B%s ne doit pas être publié', 'Identificatio referentialis singularis B%s non publicanda est');
+INSERT INTO `neg`.`datenbank_texte` (`Formular`, `Textfeld`, `de`, `gb`, `fr`, `la`) VALUES ('namenkommentar', 'NotPublicError', 'Philologisches Lemma ID N%s ist nicht zu veröffentlichen', 'Philological Lemma ID N%s is not to be published', 'Le lemme philologique ID N%s ne doit pas être publié', 'Lemma philologicum ID N%s non publicandum est');
```
mtrojan-ub: I thought that Phil. Lemma is removed from the guest view anyway, so do we really need an error message for it? Wouldn't the regular (MGH) Lemma be more proper (with M prefix instead of N)?

Does it even make sense to check whether a Lemma is public or not? A Lemma can be associated with multiple Belegforms that can be associated with different Quellen, and some of them could be public and some not. Do we assume that a Lemma is public if it is associated with at least 1 Belegform in a Quelle which is public?

juro-ub: yes the Phil Lemma (Namenkommentar) was removed from the view but the page still exist in the Source Code, so i implemented it there as well. Should i remove it ? (and the db entries as well)

The regular Lemma has the prefix M.

To determine if the regular Lemma is public the code uses following algo:
```
for (Einzelbeleg eb : listEinzelbeleg) {
            if (eb.getQuelle() != null && eb.getQuelle().getZuVeroeffentlichen() == 1) {
                throwException = false;
                break;
            }
}
```

Should i remove this logic ? Or ask Hr. Fischer ?

mtrojan-ub: I think you can remove the Phil. Lemma page and also the related error message. I guess if the public check works like this and nobody complained yet we can keep it, but then you should add display texts for the regular (non-philological) Lemma, right?

juro-ub: I have removed the namenkommentar page as well as the related error msgs.
The text for mgh_lemma is now also in the sql update file.
I tested the error msgs and it seems to be fine now

PR #350, opened 2025-05-15
Put style attributes in css files

juro-ub:


mtrojan-ub:
That's definitely a good thing to do, but: If we ever make changes to these files, the users will still have the old files in their browser cache and they won't see the changes until they explicitly delete their cache or use e.g. CTRL+F5 to re-request them. 

Other websites solve this problem by adding a GET parameter to the URL of the css file, which is the timestamp when the file was last modified. Should we do that as well? Here's an example:
https://vufind.org/demo/

https://vufind.org/demo/themes/sandal/css/compiled.css
https://vufind.org/demo/themes/sandal/css/compiled.css?_=1742213104

juro-ub:
> That's definitely a good thing to do, but: If we ever make changes to these files, the users will still have the old files in their browser cache and they won't see the changes until they explicitly delete their cache or use e.g. CTRL+F5 to re-request them.
> 
> Other websites solve this problem by adding a GET parameter to the URL of the css file, which is the timestamp when the file was last modified. Should we do that as well? Here's an example: https://vufind.org/demo/
> 
> https://vufind.org/demo/themes/sandal/css/compiled.css https://vufind.org/demo/themes/sandal/css/compiled.css?_=1742213104

Yes thats a good idea. I have implemented this for the existing static css links.

I will also search a soultion for our dynamic css generation in the addResponseHeader Function.

```
List<String> css_list = getAdditionalCss();
        String additional_css = "";
        for (String css : css_list) {
            additional_css += "<link rel=\"stylesheet\" href=\"" + css + "\" type=\"text/css\">";
        }
        request.setAttribute("additionalCss", additional_css);
```

src/main/java/de/uni_tuebingen/ub/nppm/util/Utils.java
```diff
@@ -412,4 +417,28 @@ public static String formatDate(Date date) {
         return date != null ? DATE_FORMAT.format(date) : null;
     }
 
+    //Als Fallback wird die aktuelle Systemzeit zurückgegeben um sicherzustellen das die css datei neu geladen wird
+    public static long getLastModifiedTimestampForCSS(ServletContext context, String path) throws MalformedURLException, URISyntaxException {
```
mtrojan-ub: Maybe we should name this function a little bit differently, because this is very useful and we could use it for all kinds of assets that can be cached by the client, e.g. also for JS files and so on. A proposal would be:

getLastModifiedTimestampForLocalAsset()

src/main/java/de/uni_tuebingen/ub/nppm/util/Utils.java
```diff
@@ -412,4 +417,28 @@ public static String formatDate(Date date) {
...
+        // Frühzeitiger Ausstieg: kein Zugriff auf Ressource oder kein File-URL
+        if (resource == null || !"file".equals(resource.getProtocol())) {
+            return System.currentTimeMillis(); // Fallback: Cache-Busting sicherstellen
+        }
+
+        File file = new File(resource.toURI());
+
+        // Datei existiert nicht: trotzdem Fallback geben
+        if (!file.exists()) {
+            return System.currentTimeMillis();
```
mtrojan-ub: This is a very tolerant solution here, but i wonder if it would be better to throw an exception, else we will never see that something is wrong when testing. Or should we really be tolerant here?

src/main/webapp/gast/layout/gast_einzelbeleg.css
mtrojan-ub, line 1: If CSS files are located in the /gast/layout subfolder, it doesn't really make much sense to additionally add a gast_ prefix to the filename.

src/main/java/de/uni_tuebingen/ub/nppm/util/Utils.java
juro-ub: i have renamed the function so we can use it in the future for js

src/main/java/de/uni_tuebingen/ub/nppm/util/Utils.java
juro-ub: The functions throws now an Exception on failure. Should the Message in the Exception should also be translated ? It is now only in english

src/main/webapp/gast/layout/gast_einzelbeleg.css
juro-ub: yes, i have removed the gast_ prefix

src/main/java/de/uni_tuebingen/ub/nppm/util/Utils.java
mtrojan-ub: Maybe this is not necessary => this is a server sided error that only developers are able to solve and there is something fundamentally wrong (compared to other messages where the user should know that e.g. he requested a non-existing ID)

src/main/webapp/gast/layout/gast_stat.css
mtrojan-ub, line 1: the gast_ prefix should also be removed from gast_stat.css

juro-ub: yes i have removed it

mtrojan-ub:
Looks good now, just one more thing: When using the browser's dev console it seems like the change is active for e.g. einzelbeleg.css, but not for lots of basic assets like e.g. layout.css, jquery-ui.min.css, merged.css and so on. Could you please use the mechanism there as well?

Maybe also for the other assets, especially JS files (only if it's not too complicated).

grafik

juro-ub:
> Looks good now, just one more thing: When using the browser's dev console it seems like the change is active for e.g. einzelbeleg.css, but not for lots of basic assets like e.g. layout.css, jquery-ui.min.css, merged.css and so on. Could you please use the mechanism there as well?
> 
> Maybe also for the other assets, especially JS files (only if it's not too complicated).
> 
> grafik

i have create the versioned href now for all css and js includes.
I tested the backend and frontend pages and it looks fine

mtrojan-ub:
Looks good now, thanks :+1:

PR #351, opened 2025-05-15
Fix Menu Bar Width

juro-ub:


src/main/webapp/gast/layout/layout.css
```diff
@@ -211,3 +211,12 @@ Um Seiteneffekte zu ermeiden lasse ich das so und setze die neue Breite hier mit
 #erweiterte-suche #tab-1 .ut-table__row > td:first-child {
     width: 370px !important;
 }
+
+.menuBarFitcontent{
+    min-width: 1200px !important;
```
mtrojan-ub: We need a solution that also works with all smaller / mobile resolutions, so we definitely cannot assume 1200px as min-width!

juro-ub: I have removed the min-width attribute. The `"width: fit-content"` is enough to display the menubar correctly. i have tested it also in the mobile view.

The problem was, that the menu bar in some languages is to small for the menu-items

mtrojan-ub:
Looks good for the desktop version, but in the mobile version it looks like this:
grafik

Compared to uni-tuebingen.de => without input field but with regular sizes:
grafik

- The menu entries are very small and take only a small part of the available width
- The icon `>` on the right side is completely missing
- the input field doesn't seem to be really part of the entry (the underline is only there for the ID label but not for the input field).

juro-ub:
> Looks good for the desktop version, but in the mobile version it looks like this: grafik
> 
> Compared to uni-tuebingen.de => without input field but with regular sizes: grafik
> 
> * The menu entries are very small and take only a small part of the available width
> * The icon `>` on the right side is completely missing
> * the input field doesn't seem to be really part of the entry (the underline is only there for the ID label but not for the input field).

1. The Menu Bar has now full width of the mobile screen
2. The '>' item only appears if there is a sub-menu-entry. since we have only a top-level-menu with no sub items, it is not displayed.
3. I have made a fix for the underline

PR #359, opened 2025-05-30
Create Constant forbiddenLemmaSubstring

juro-ub:
Create Constant forbiddenLemmaSubstring to exclude lemmas in frontend


src/main/java/de/uni_tuebingen/ub/nppm/db/EinzelbelegDB.java
```diff
@@ -224,13 +225,13 @@ public static List<EinzelbelegHatStand> getListEinzelbelegHatStand(int einzelbel
...
             /*
-                Exclude Einzelbelege that are linked to a MGHLemma which contains [???] in Frontend
+                Exclude Einzelbelege that are linked to a MGHLemma which Constants.forbiddenLemmaSubstring in Frontend
             */
             String sql = "SELECT e.ID "
                     + "FROM einzelbeleg e "
                     + "  JOIN quelle q ON e.QuelleID = q.ID AND q.zuVeroeffentlichen = 1 "
                     + "  LEFT JOIN einzelbeleg_hatmghlemma eh ON eh.EinzelbelegID = e.ID "
-                    + "  LEFT JOIN mgh_lemma m ON m.ID = eh.MGHLemmaID AND m.MGHLemma LIKE '%[???]%' "
+                    + "  LEFT JOIN mgh_lemma m ON m.ID = eh.MGHLemmaID AND m.MGHLemma LIKE '%"+Constants.forbiddenLemmaSubstring+"%' "
```
mtrojan-ub: Whenever NativeQuery is used, a proper escaping function must be called (even if Constants.forbiddenLemmaSubstring does not contain dangerous characters right now).

juro-ub: Yes, i have used the exiting function DBtoDB to escape these String.

src/main/java/de/uni_tuebingen/ub/nppm/util/Utils.java
```diff
@@ -34,6 +34,15 @@ public static String safeToString(Object o, String fallback) {
         }
     }
 
+    public static String DBtoDB(String s) {
```
mtrojan-ub: - Should we move this function to the DB namespace? (AbstractBase)
- We should rename this either to escapeDB (if it stays here) or just escape (if moved to db.AbstractBase)
- Whether `'` or `"` needs to be escaped depends on which of these characters is used by the caller, so the caller should pass the quote character as second parameter

juro-ub: i have renamed the function to `escape(...)` and moved it to AbstractBase.

The escape functions uses a the quote character as a parameter now

src/main/java/de/uni_tuebingen/ub/nppm/db/LemmaDB.java
```diff
@@ -46,7 +48,7 @@ public static Integer getNextPublicMGHLemmaID(int id) throws Exception {
                     + "      JOIN einzelbeleg e ON e.ID = h.EinzelbelegID "
                     + "      JOIN quelle q ON e.QuelleID = q.ID "
                     + "      WHERE q.ZuVeroeffentlichen = 1 AND mgh_lemma.ID = :id "
-                    + "        AND mgh_lemma.MGHLemma NOT LIKE '%[???]%' "
+                    + "        AND mgh_lemma.MGHLemma NOT LIKE '%"+AbstractBase.escape(Constants.forbiddenLemmaSubstring,'\'','"')+"%' "
```
mtrojan-ub: When building the MySQL Query, the string will look like this:

`... AND mgh_lemma.MGHLemma NOT LIKE '%[???]%' ...`

So the used delimiter is a single quote `'`, which means that the call to the escape function should only pass the single quote as parameter and not the double quote. (This also affects the other cases).

juro-ub: Yes i have removed it in all cases

juro-ub:
Merge conflicts are solved now

PR #366, opened 2025-06-05
Fallback lang eng

juro-ub:
Set Default Language to gb

### DIFF
--- a/src/main/java/de/uni_tuebingen/ub/nppm/servlet/gast/SharedHtmlServlet.java
+++ b/src/main/java/de/uni_tuebingen/ub/nppm/servlet/gast/SharedHtmlServlet.java
@@ -5,6 +5,7 @@ import de.uni_tuebingen.ub.nppm.model.Content;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.PrintWriter;
+import javax.persistence.NoResultException;
 
 public class SharedHtmlServlet extends AbstractGastServlet {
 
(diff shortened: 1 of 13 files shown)


src/main/webapp/suche/freie_suche.jsp
```diff
@@ -320,7 +320,7 @@
 
   // ######### SUCHANFRAGE ##########
 
-    String sprache = "de";
+    String sprache = "gb";
```
mtrojan-ub: `gb` should not be hardcoded in all of these cases => `Constants.DEFAULT_LANG` should be used whenever possible

src/main/java/de/uni_tuebingen/ub/nppm/servlet/gast/SharedHtmlServlet.java
```diff
@@ -44,25 +45,30 @@ protected void generatePage(HttpServletRequest request, HttpServletResponse resp
...
-            Content content = ContentDB.getByNameAndLanguage(myFile, "de");
-
+        } catch (NoResultException e) {
+            Content content = null;
+            try{
+                //Versuche erst englisch zu holen
+                content = ContentDB.getByNameAndLanguage(myFile, "gb");
+            }catch(NoResultException e1){
+                // Fehlerbehandlung benutze Standard Sprache Deutsch, da für nicht alle wie Hilfe.html eine Englische Version vorhanden ist
+                content = ContentDB.getByNameAndLanguage(myFile, "de");
```
mtrojan-ub: There should be a constant defined for `de` as well, e.g. `Constants.FALLBACK_LANG`

src/main/webapp/suche/freie_suche.jsp
juro-ub: I have created and used the Constant DEFAULT_LANG now

src/main/java/de/uni_tuebingen/ub/nppm/servlet/gast/SharedHtmlServlet.java
juro-ub: Yes also Fallback Language is defined in a constant now

mtrojan-ub:
This PR seems to be related to #370 to a certain degree. I think the other PR can't be merged yet anyway because there are display texts missing from our stakeholders. Should we merge this one first & re-use the constants for default / fallback language in the other PR?

juro-ub:
> This PR seems to be related to #370 to a certain degree. I think the other PR can't be merged yet anyway because there are display texts missing from our stakeholders. Should we merge this one first & re-use the constants for default / fallback language in the other PR?

Yes we can merge this PR first and then reuse the constants in the other PR

PR #379, opened 2025-06-25
CSV Export Einfache Suche

### LINKED ISSUE #375: CSV-Export / Einfaches Suchergebnis
Noch wichtig bis Mitte 2025

juro-ub:


src/main/java/de/uni_tuebingen/ub/nppm/servlet/gast/ExportCsvServlet.java
```diff
@@ -0,0 +1,56 @@
...
+
+        try (PrintWriter out = response.getWriter()) {
+            // CSV-Kopfzeile
+            out.println(String.join(";", fieldNames));
+
+            for (Map row : result) {
+                List<String> values = new ArrayList<>();
+                for (String field : fieldNames) {
+                    Object val = row.get(field);
+                    String text = val == null ? "" : val.toString().replace("\"", "\"\"");
```
mtrojan-ub: Doing this kind of escaping by yourself is usually tricky, because there can be certain edge-cases like e.g. newlines inside a single field that can break the content.

Can we use a library instead to generate the file, e.g. OpenCSV?
https://www.geeksforgeeks.org/java/writing-a-csv-file-in-java-using-opencsv/

src/main/java/de/uni_tuebingen/ub/nppm/servlet/gast/ExportCsvServlet.java
```diff
@@ -0,0 +1,56 @@
...
+        try {
+            result = SucheDB.getEinfacheSucheResult(query);
+        } catch (Exception e) {
+            response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR, "Failed to retrieve search results: " + e.getMessage());
+            return;
+        }
+
+        try (PrintWriter out = response.getWriter()) {
+            // CSV-Kopfzeile
+            out.println(String.join(";", fieldNames));
```
mtrojan-ub: Maybe we should translate the header entries?

src/main/java/de/uni_tuebingen/ub/nppm/servlet/gast/ExportCsvServlet.java
juro-ub: i have used opencsv now

src/main/java/de/uni_tuebingen/ub/nppm/servlet/gast/ExportCsvServlet.java
juro-ub: the translation is now the same as in the einfachesuche servlet

pom.xml
```diff
@@ -128,6 +128,11 @@
             <artifactId>gson</artifactId>
             <version>2.13.1</version>
         </dependency>
+        <dependency>
+            <groupId>com.opencsv</groupId>
+            <artifactId>opencsv</artifactId>
+            <version>5.9</version>
```
mtrojan-ub: Version 5.9 is from 2023, please use the latest version 5.11.2, see also:
https://mvnrepository.com/artifact/com.opencsv/opencsv

src/main/webapp/gast/suche/einfache_suche.jsp
```diff
@@ -13,7 +14,9 @@
 <%@ page import="java.io.*" isThreadSafe="false" %>
 <%@ page import="java.awt.Color" isThreadSafe="false" %>
 
-
+<a href="<%= Utils.getBaseUrl(request) %>/gast/export_csv?query=<%= URLEncoder.encode(request.getParameter("query"), "UTF-8") %>" class="ut-btn">
+    Export als CSV
```
mtrojan-ub: Please add a translated display text for this

src/main/java/de/uni_tuebingen/ub/nppm/servlet/gast/ExportCsvServlet.java
```diff
@@ -0,0 +1,87 @@
...
+
+    protected void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        String query = request.getParameter("query");
+        if (query == null || query.trim().length() < 3) {
+            response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Invalid Query");
+            return;
+        }
+
+        response.setContentType("text/csv; charset=UTF-8");
+        String filename = "suchergebnisse.csv";
```
mtrojan-ub: Is there also an already existing translated display text that we can use for the filename?

juro-ub: yes i have used now the titel from suchergebnis:
`Language.getTextfield(request.getSession(), "such_ergebnis", "Titel")`

src/main/webapp/gast/suche/einfache_suche.jsp
juro-ub: yes added!

pom.xml
juro-ub: version is now the latest

mtrojan-ub:
Thanks! This is a good first draft. I think that there are still some changes necessary (e.g. i don't think that the end user will know e.g what "Q von J." or "EB von JH." means), but the question is whether we change the headlines or provide any kind of explanation in the frontend... however this is a good draft to discuss further steps with Mr Patzold, so i will merge this & deploy it on the test server.

PR #382, opened 2025-06-27
Custom Privacy Page

mtrojan-ub:


mtrojan-ub:
For now, the custom privacy page will be hidden, but we need to roll it out in the test environment to allow proofreading, so i will merge it at the current state & add corrections later if necessary.

PR #384, opened 2025-07-02
Fix GND Link

juro-ub:
Problem: Der Name des Datenfeldes hat sich mit einem früheren SQL Update geändert

### DIFF
--- a/src/main/webapp/gast/person.jsp
+++ b/src/main/webapp/gast/person.jsp
@@ -91,7 +91,7 @@
                     <jsp:include page="../inc.erzeugeFormular.jsp">
                         <jsp:param name="ID" value="<%=id%>" />
                         <jsp:param name="Formular" value="person" />
-                        <jsp:param name="Datenfeld" value="GNDLink" />
+                        <jsp:param name="Datenfeld" value="IconLink" />
                     </jsp:include>
                      &nbsp;
                     <jsp:include page="../inc.erzeugeFormular.jsp">
--- a/src/main/webapp/gast/person.jsp
+++ b/src/main/webapp/gast/person.jsp
@@ -91,7 +91,7 @@
                     <jsp:include page="../inc.erzeugeFormular.jsp">
                         <jsp:param name="ID" value="<%=id%>" />
                         <jsp:param name="Formular" value="person" />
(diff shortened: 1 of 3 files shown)


src/main/webapp/gast/person.jsp
```diff
@@ -91,7 +91,7 @@
                     <jsp:include page="../inc.erzeugeFormular.jsp">
                         <jsp:param name="ID" value="<%=id%>" />
                         <jsp:param name="Formular" value="person" />
-                        <jsp:param name="Datenfeld" value="GNDLink" />
+                        <jsp:param name="Datenfeld" value="IconLink" />
```
mtrojan-ub: Looks like noarray.iconlink.jsp handles GND as well as Wikidata Links. Before your change, there are 2 different kinds of jsp:includes to inc.erzeugeFormular.jsp => one for GNDLink and the other one for WikidataLink. If you change the GNDLink to IconLink, the WikidataLink remains, which is kind of confusing.

Do we need to merge them together? Or do we rather need to clean up the database so that both of the existing calls get redirected correctly?

juro-ub: There are two database entries in datenbank_mapping. one for gnd and on for wikidata. Because you need two different feldtypes and Zielattirbutes and zieltabellen. So i think you can not merge these database entries together.

```
Formular, Datenfeld, Feldtyp, ZielAttribut 
____________________________________________
'person', 'IconLink', 'gndlink', 'GND' 
'person', 'WikidataLink', 'wikidatalink', 'Wikidata'
```

And to handle these feldtypes the the jsp file noarray.iconlink.jsp exists.

So now there are two databaseentries and only one include for these (noarray.iconlink).

One Option whould be two rename the Datenfeld "IconLink" in datenbank_mapping back to "GNDLink". For better understanding

mtrojan-ub: That sounds good! Can you please rename it and change the references accordingly if necessary? (and also test whether changing the values in the backend still works afterwards)

juro-ub: I have changed the Datenfeld name to GNDLink. It works now in the 
Backend:
image
and Frontend:
image

PR #388, opened 2025-07-09
Refactor Field Quellengattung from Einzelbeleg to Quelle

### LINKED ISSUE #372: Feld Quellengattung vom Einzelbeleg zur Quelle schieben
+ Liste erstellen für Quellen deren Einzelbelege mehr als 1 Gattungs-Angabe haben

juro-ub:
Issue https://github.com/ubtue/neg/issues/372


mtrojan-ub:
Is this still a Draft or Ready for review?

juro-ub:
> Is this still a Draft or Ready for review?

The PR is on draft because the requirement for the update script is, that each Quelle has only one Gattung.
See https://github.com/ubtue/neg/issues/372

mtrojan-ub:
So if i understand correctly that means it's ready for review but not ready for merge :smirk: I will have a look when time permits and also try to test it / leave some comments if necessary.

juro-ub:
> So if i understand correctly that means it's ready for review but not ready for merge 😏 I will have a look when time permits and also try to test it / leave some comments if necessary.

Yes exactly

juro-ub:
Migration script completed.
It is important that the file /var/lib/mysql-files/quelle_with_multiple_gattungen-2.csv exists when executing the SQL script. It is currently located in the update/ folder.

updates/quelle_with_multiple_gattungen-2.csv
mtrojan-ub, line 1: This file as well as the SQL update should be moved to the db/cleanup directory

juro-ub: Moved!

mtrojan-ub:
webapp/einzelbeleg.jsp doesn't seem to be on the list of changed files. Shouldn't the old field be removed there as well?

db/cleanup/quelle_with_multiple_gattungen.sql
```diff
@@ -0,0 +1,43 @@
...
+JOIN selektion_quellengattung ON selektion_quellengattung.Bezeichnung = bereinigung_csv.gattungen_NEU
+SET einzelbeleg.QuelleGattungID = selektion_quellengattung.ID;
+
+DROP TABLE bereinigung_csv;
+
+ALTER TABLE quelle ADD COLUMN QuelleGattungID INT NOT NULL DEFAULT '-1';
+ALTER TABLE quelle ADD CONSTRAINT fk_quelle_quellengattungid FOREIGN KEY (QuelleGattungID) REFERENCES selektion_quellengattung(ID) ON UPDATE CASCADE;
+
+#optional
+#ALTER TABLE einzelbeleg DROP COLUMN QuelleGattungID;
```
mtrojan-ub: And this should also be enabled, but at the very end of this file after the migration was executed

juro-ub: yes, fixed!

juro-ub:
> webapp/einzelbeleg.jsp doesn't seem to be on the list of changed files. Shouldn't the old field be removed there as well?

yes it is removed now

mtrojan-ub:
Two more points:
- The old entry in datenbank_mapping for einzelbeleg.Quellengattung should be removed
- In the search mask it looks like you can no longer select the values `-` or `?` like in the old version, so the default right now is "Annalen" which means you can no longer do an empty search

juro-ub:
> Two more points:
> 
> * The old entry in datenbank_mapping for einzelbeleg.Quellengattung should be removed
> * In the search mask it looks like you can no longer select the values `-` or `?` like in the old version, so the default right now is "Annalen" which means you can no longer do an empty search

- Old entry is removed now
- '-' and '?' are in the list now
image

PR #391, opened 2025-07-18
Fix freie Suche

### LINKED ISSUE #363: Erweiterte Suche: Einschränkung nach mehreren Quellen erlauben
Ist als Wunsch entstanden, da es einen Unterbestand verschiedener Verbrüderungsbücher gibt in denen man gemeinsam suchen können soll. Diese sollen potenziell auch auf die MGH verlinkt werden, allerdings ist noch offen ob die URLs dem selben Schema entsprechen (Herr Patzold fragt die MGH nach Beispielen).

(Die Umsetzung soll ggf. erst nach Pfingsten erfolgen, daher weise ich eher @juro-ub als @rnaiser zu.)

juro-ub:
Fix for the problem that the user was not able to search for a specific source, as discussed in the meeting on 17.7.25

- use escapeSQL
- fix for Parameter "Quelle"

### DIFF
--- a/src/main/java/de/uni_tuebingen/ub/nppm/db/AbstractBase.java
+++ b/src/main/java/de/uni_tuebingen/ub/nppm/db/AbstractBase.java
@@ -26,11 +26,11 @@ import javax.persistence.Table;
 import org.hibernate.type.StringType;
 
 public class AbstractBase {
-    public static Map<Character, String> sqlEscapes = new HashMap<>();
+    public static Map<Character, String> sqlEscapesSingleQuotes = new HashMap<>();
 
     static {
-        sqlEscapes.put('\'', "''");
-        sqlEscapes.put('\\', "\\\\");
+        sqlEscapesSingleQuotes.put('\'', "''");
+        sqlEscapesSingleQuotes.put('\\', "\\\\");
     }
     protected static SessionFactory sessionFactory;
 
--- a/src/main/java/de/uni_tuebingen/ub/nppm/db/AbstractBase.java
+++ b/src/main/java/de/uni_tuebingen/ub/nppm/db/AbstractBase.java
@@ -505,14 +505,11 @@ public class AbstractBase {
     public static String escape(String s, char... delimiters) {
         if (s != null) {
             s = s.replace("\\", "\\\\");
+            s = s.replace("'", "''");
             for (char delimiter : delimiters) {
                 s = s.replace(String.valueOf(delimiter), "\\" + delimiter);
             }
         }
         return s;
-    }
-
-    public static String escapeSql(String s) {
-        return s == null ? null : s.replace("'", "''");
     }
 }
--- a/src/main/java/de/uni_tuebingen/ub/nppm/db/AbstractBase.java
+++ b/src/main/java/de/uni_tuebingen/ub/nppm/db/AbstractBase.java
@@ -30,7 +30,6 @@ public class AbstractBase {
 
     static {
         sqlEscapesSingleQuotes.put('\'', "''");
-        sqlEscapesSingleQuotes.put('\\', "\\\\");
     }
     protected static SessionFactory sessionFactory;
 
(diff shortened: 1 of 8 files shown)


src/main/java/de/uni_tuebingen/ub/nppm/db/AbstractBase.java
```diff
@@ -511,4 +511,8 @@ public static String escape(String s, char... delimiters) {
         }
         return s;
     }
+
+    public static String escapeSql(String s) {
```
mtrojan-ub: There is already a escape function in AbstractBase which should be used

juro-ub: i have changed it to the existing escape function

src/main/java/de/uni_tuebingen/ub/nppm/db/AbstractBase.java
```diff
@@ -505,6 +505,7 @@ public static String getProvenanceSource(String id, String tabelle) throws Excep
     public static String escape(String s, char... delimiters) {
         if (s != null) {
             s = s.replace("\\", "\\\\");
+            s = s.replace("'", "''");
```
mtrojan-ub: This must not be hardcoded. The caller must decide whether ' should be passed as delimiter or not.

juro-ub: I changed the function. it uses now as a parameter a map.
For sqlEscaped it looks like this:
```
    public static Map<Character, String> sqlEscapes = new HashMap<>();

    static {
        sqlEscapes.put('\'', "''");
        sqlEscapes.put('\\', "\\\\");
    }
```
This is necessary because not all characters are escaped with a backslash. For example ' becomes '' in SQL.

src/main/java/de/uni_tuebingen/ub/nppm/db/AbstractBase.java
```diff
@@ -26,7 +26,12 @@
 import org.hibernate.type.StringType;
 
 public class AbstractBase {
+    public static Map<Character, String> sqlEscapes = new HashMap<>();
 
+    static {
+        sqlEscapes.put('\'', "''");
```
mtrojan-ub: Again: We cannot define any type of single or double quotes to be used for escaping in all queries, because it depends on how the query is built by the caller.

Here are some examples, e.g.
`INSERT INTO selektion_verwandtschaftsgrad (Bezeichnung) VALUES (<placeholder>);`
| Number | `<placeholder>`    | Inserted Value |
| -------- |-------- | ------- |
| 1 | `"'test'"` | `'test'` |
| 2 | `'"test"'` | `"test"` |
| 3 | `'\'test\''` | `'test'` |
| 4 | `"\"test\""` | `"test"` |
| 5 | `"\'test\'"` | `\'test\'` |
| 6 | `'\"test\"'` | `\"test\"` |


Cases 1-4 are usually what you want, but you definitely want to avoid cases 5-6. That's why you need to let the escape function know which kind of quotes you used in every specific case.

juro-ub: The escape function takes an escape map as parameter, so the caller decides which characters to escape and how. So you can define escaping for single quotes and double quotes.

Currently i only have the case to escape single quotes. So i defined a global map for singlequotes now:
```
    public static Map<Character, String> sqlEscapesSingleQuotes = new HashMap<>();

    static {
        sqlEscapesSingleQuotes.put('\'', "''");
        sqlEscapesSingleQuotes.put('\\', "\\\\");
    }
```
    
    so the caller can use this to do the correct escaping. For example 
`"mgh_lemma.MGHLemma NOT LIKE '%"+AbstractBase.escape(Constants.forbiddenLemmaSubstring,AbstractBase.sqlEscapesSingleQuotes)+"%' "`

And if he wants to escape double quotes he must use a different map as parameter for the escape function.

Is it correct like this ? Or do i still have an understanding mistake ?

mtrojan-ub: Seems like you're right about the fact that e.g. a single quote can be escaped with another single quote if it is also the delimiter of the query, e.g. if you do an insert of the String `''''` (4 single quotes), just one of them will be inserted as string. I thought this would only be the case in PostgreSQL, but it also appears to be the case for MySQL.

However, i cannot think of any case whatsoever when we don't want to escape the backslash, so i like the idea more how it was before that the backslash is hardcoded within the function and does not need to be passed via parameter, whether array or not. I also like the old mechanism that you can just pass additional parameters instead of an array.

So, if you insist to do it like this, i would suggest that we just have 2 implementations of the escape function (overloading) so you can pass either `char... delimiters` or `Map<Character, String> escapeMap`, but the implementation of one should use the other, and also the `\` should be hardcoded anyway. Is that OK for you?

juro-ub: yes! fine for me.
I have overloaded the escape function. One function with `char... delimiters`
and one functions with a `custom map`

So we can keep the old behaviour and have an additional behaviour for user defined escaping.

PR #392, opened 2025-07-21
Performance Optimizations

mtrojan-ub:


juro-ub:
While build the application i receive error like:

```
testList  Time elapsed: 0.088 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: Unable to create requested service [org.hibernate.cache.spi.CacheImplementor]
	at de.uni_tuebingen.ub.nppm.controller.test.SucheDBTest.testList(SucheDBTest.java:23)
```

Is it the same in your case ?

mtrojan-ub:
You're right... i usually don't run the tests when building locally every time (cause it takes lots of additional time to run them). Do you know whether the settings for the second level cache need to be separately included for the test environment somehow?

/edit: Seems like it works if i remove the following line, so there might be a classpath-related problem with junit
`settings.put("hibernate.javax.cache.uri", "classpath:ehcache.xml");`

mtrojan-ub:
The testing problem is solved now

PR #396, opened 2025-07-25
Fix Generate Url

juro-ub:


src/main/webapp/gast/quelle.jsp
```diff
@@ -134,8 +134,54 @@
...
+<%
+    String paramName = URLEncoder.encode("Quellenliste[]", "UTF-8");
+    String baseUrl = Utils.getBaseUrl(request);
+    String idParam = String.valueOf(id);
+
+    String link = baseUrl + "/gast/suchergebnis"
+        + "?einzelbelegeVonQuelle=true"
+        + "&" + paramName + "=" + idParam
+        + "&form=freie_suche"
+        + "&NeGID="
```
mtrojan-ub: Can we remove all parameters that only contain empty or default values?

juro-ub: Fixed now!

PR #398, opened 2025-07-25
Initial cronjob for Sitemap

mtrojan-ub:
Notes/TODO:
- [x] Generate Sitemap on Servers via cronjob
- [x] Split up sitemap into several pieces per entity 
  - [x] Make sure no part is bigger than 10MB
- [x] Add base sitemap with static pages
- [x] Ensure all listed entities are ZuVeroeffentlichen
- [x] Introduce more generic interfaces for getZuVeroeffentlichen() + getId() (maybe even getPersistentIdentifier())
- [x] Ensure symlink to data dir doesn't get replaced during WAR build
- [x] Ensure Timestamps are in correct format + unified
- [x] Register Sitemap + /id/ resolver in robots.txt

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ For servers (ZDV):
 - cronjobs
     - See "cronjobs" file in the root directory, add entries to crontab
     - DATA_DIR and LOG_DIR should be created manually
+    - Also make sure DATA_DIR/sitemaps is symlinked in your tomcat/webapps directory
     - Create /root/.neg.properties (see above)
 - Firewall
     - adjust firewall scripts, see /zdv-system/scripts/ipt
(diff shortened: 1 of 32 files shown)


juro-ub (commented):
Worked fine.
Maybe one point. The sitemap-einzelbelege.xml has over 399474 entries.
Is it allowed sitemaps of this size ?
My first google search says it is limited to 50.000 urls

mtrojan-ub:
That's right, and also there might be other bots who have problems with sitemaps that are > 10 MB so we need to split them up. I already had a TODO at the top of this PR, together with some other open changes.

mtrojan-ub:
This might still need some optimization, e.g. to set changefreq if a lastmod date cannot be determined. However, it might be better to offer a sitemap and improve them later since the basic URL list seems to be fine, so we will merge this as-is right now and might do further improvements later.

